PR #2, opened 2015-05-25
add check-load

tatsuru:
add loadavg check like [Nagios Plugin](https://github.com/nagios-plugins/nagios-plugins/blob/master/plugins/check_load.c)


Songmu:
LGTM

PR #27, opened 2015-11-10
bug check-log return int64

matsuzj:
strconv.Atoi return of the int32.
Changing to return to int64.

Error Message
```
/usr/local/bin/check-log --file=logfile --pattern='paten'
LOG UNKNOWN: strconv.ParseInt: parsing "2203473635": value out of range
```


Songmu:
Thank you! :bow:

PR #35, opened 2015-11-23
Fix bugs of check-log

Songmu:
- care file rotation properly
- fix logfile handling when the file is not ended with newline character.


itchyny:
LGTM!

PR #41, opened 2015-12-04
Check tcp add ipv6 test

tkuchiki:
- Add IPv6 test
- strings.Split -> net.SplitHostPort
- remove binary
- Update README


Songmu:
adjust and merged at #42 

I didn't know `addons.apt` section.
Thank you!!!

PR #49, opened 2015-12-15
add retry options to check-procs

etsxxx:
I get CRITICAL error from `chec-procs` when logrotate, reload process or sometimes.
I added **'retry'** feature.
The user can use this feature by adding option `--retry=NUM` and change retry interval by `--interval=SECONDS`.

Examples:
No vim processes found, get CRITICAL.

```
$ ./check-procs -C 1 -p "vim"
Procs CRITICAL: Found 0 matching processes; cmd /vim/
```

with `--retry=2` and `--interval=5`, someone runs vim within 5 seconds, get OK.

```
$ ./check-procs --retry=2 --interval=5 -C 1 -p "vim"
### someone runs vim within 5 seconds
Procs OK: Found 1 matching processes; cmd /vim/
```

Regards.

### DIFF
--- a/check-procs/check_procs.go
+++ b/check-procs/check_procs.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"regexp"
 	"strconv"
+	"time"
 
 	"github.com/jessevdk/go-flags"
 	"github.com/mackerelio/checkers"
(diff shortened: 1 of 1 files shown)


Songmu:
Thank you for your pull request.

But, I think plugin itself should not care retrying. It should be
cared in other layer.

If once a plugin have retryable interface, all of plugins going to have to
be implemented retryable interface too. It is nonsense.

In such case, you can write some wrapper script like following.

```
#!/bin/bash
for i in 1 2 3; do
  "$@"
  ret=$?
  [[ $ret = 0 ]] && break
  sleep 5
done
exit $ret
```

And use this shell script as following.

    ./retry.sh ./check-procs -C 1 -p "vim"

How do you think?

Songmu:
And we Mackerel develpers plan to implement `max_check_attempts` feature at server side for sensitive checkers. Please wait them :bow:

etsxxx:
> If once a plugin have retryable interface, all of plugins going to have to
be implemented retryable interface too.

Yes, you're right.
And the serverside features are welcome. It's a good solution.

I comment to 'retry.sh'.
If you use 'wrapper' with mackerel-agent commands, you should care double quotations, single quotations and escapes.
If you use BASH to wrapper, it's like a hell.
So I welcome serverside features.

I'm looking forward to it, thank you.

Songmu:
> If you use 'wrapper' with mackerel-agent commands, you should care double quotations, single quotations and escapes.

No. This `retry.sh` doesn't accept strings, but arguments. So escapings are not required.

There is `retry` command like following.
http://suzuken.hatenablog.jp/entry/2015/08/26/195708

etsxxx:
Oh, I missed double quotes of `"$@"` in testing, you're right. `retry.sh` works.
But wrappers counts itself, so I should +1 `-C` and `-W` options to check correctly.

Thank you to telling knowledge.

PR #53, opened 2016-01-05
add check-mailq, currently only available for postfix.

tnmt:
inspired by [check_mailq.pl](https://github.com/nagios-plugins/nagios-plugins/blob/master/plugins-scripts/check_mailq.pl)

Nagios plugin covers several MTA's check by one script, so I started to write like that. also this plugin currently works for only Postfix...

Is it good to separate MTA check into each one script?

### DIFF
--- a/check-mailq/README.md
+++ b/check-mailq/README.md
@@ -1,0 +1,22 @@
+# check-mailq
+
+## Description
+
+Monitor mail queue count.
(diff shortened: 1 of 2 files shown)


tnmt:
sorry, it's wrong branch to fork. I will try to make other p-r.

PR #71, opened 2016-02-09
added check-file-size

hiroakis:
The plugin checks the files size of specified directory.
As an use case, to check the Fluentd files buffer directory if the files buffer increases large size.


hiroakis:
It seems that testing of check-tcp fails. I don't know why.

hiroakis:
I opened a new pull request since check-tcp testing has faced into unknown failure. So, I close this pull request.

following is a new pull request
https://github.com/mackerelio/go-check-plugins/pull/72

PR #76, opened 2016-02-19
Fixed slurp. conn.read returns content with EOF

kazeburo:
Some time check-tcp could not retrieve content with EOF error.

net.read returns content with EOF error.
ref. http://dsas.blog.klab.org/archives/cat_50043510.html


Songmu:
Thank you!

PR #77, opened 2016-02-22
Fix check-tcp IPv6 testcase on OSX(?)

hanazuki:
The IPv6 test for check-tcp fails on my OS X machine for some reason. This patch fixes it, at least on my environment.

I don't know why the original code doesn't run on my OS X environment but I see no reason using Hijacker: http.HandleFunc handles everything fine.


Songmu:
Thank you!

PR #87, opened 2016-03-13
Fix setting example of check-uptime README

a-know:
About options for warning threshold, written in `Setting` use `--warning-under=600`, but it seems correct to use `--warn-under=600` seemingly.
(Other plugins are used `--warn-under`, too.)


Songmu:
Oh, it's my mistake... I prefer `--warning-under` to `--warn-under`, but implementation is not.

a-know:
Thank you for your reply!
I opened #88, please review that.

PR #88, opened 2016-03-16
Use `--warning-*` instead of `--warn-*` 

a-know:
relate with https://github.com/mackerelio/go-check-plugins/pull/87#issuecomment-197206911

@Songmu intend to use `--warning-under` `--warning-over`, but implementation is not.
This is the pull request to amend that implementation.

### DIFF
--- a/check-uptime/README.md
+++ b/check-uptime/README.md
@@ -14,8 +14,8 @@ command = "/path/to/check-uptime --warning-under=600 --critical-under=120"
 ## Options
 
 ```
-  -w, --warn-under=N        Trigger a warning if under the seconds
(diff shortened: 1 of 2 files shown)


a-know:
I worried about amend `go-check-plugins/check-procs` or not.
(`go-check-plugins/check-procs` used `--warn-*`, too. https://github.com/mackerelio/go-check-plugins/blob/master/check-procs/check_procs.go#L15)

If it is better to fix `go-check-plugins/check-procs`, I can also fix that in this PR.

Songmu:
Sorry for slow response.
I refine this patch for maintaining backward compatibility at #104.

Thank you!

PR #100, opened 2016-03-28
Add packaging files for Amazon Linux

stanaka:


stanaka:
The issue of supporting Amazon Linux has been solved with another solution.

PR #113, opened 2016-06-13
Fixed argument parser error:

karupanerura:
This error is always occurred (but, not printed):
`boolean flag `-r/percpu' may not have default values, they always default to `false' and can only be turned on`


Songmu:
Thank you!

PR #115, opened 2016-08-07
Add option for skip searching logfile content if logfile does not exist

a-know:
`check-log` exit `UNKNOWN` if logfile which specified by `LogFile` does not exist.
This pull request is add `skip-not-exist` option for skip searching logfile content if logfile does not exist.

### DIFF
--- a/check-log/check-log.go
+++ b/check-log/check-log.go
@@ -29,7 +29,7 @@ type logOpts struct {
 	CaseInsensitive bool    `short:"i" long:"icase" description:"Run a case insensitive match"`
 	StateDir        string  `short:"s" long:"state-dir" default:"/var/mackerel-cache/check-log" value-name:"DIR" description:"Dir to keep state files under"`
 	NoState         bool    `long:"no-state" description:"Don't use state file and read whole logs"`
-	Missing         string  `long:"missing" default:"" value-name:"(WARNING|OK)" description:"Exit status when log files missing"`
+	Missing         string  `long:"missing" default:"" value-name:"(CRITICAL|WARNING|OK|UNKNOWN)" description:"Exit status when log files missing"`
 	patternReg      *regexp.Regexp
 	excludeReg      *regexp.Regexp
 	fileList        []string
@@ -100,7 +100,7 @@ func regCompileWithCase(ptn string, caseInsensitive bool) (*regexp.Regexp, error
 
 func validateMissing(missing string) bool {
 	switch missing {
-	case "WARNING", "OK", "":
+	case "CRITICAL", "WARNING", "OK", "UNKNOWN", "":
 		return true
 	default:
 		return false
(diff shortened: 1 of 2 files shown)


Songmu:
How about `--skip-missing` or `--missing=(WARNING|OK)` ?

a-know:
I think `--missing=(WARNING|OK)` is good so I fixed it.

check-log/check-log.go
```diff
@@ -154,6 +164,9 @@ func (opts *logOpts) searchLog(logFile string) (int64, int64, string, error) {
 
 	f, err := os.Open(logFile)
 	if err != nil {
```
Songmu: It is not always file missing error. You should `continue` immediately when file is missing outsisde this function.

check-log/check-log.go
```diff
@@ -113,9 +114,14 @@ func run(args []string) *checkers.Checker {
 
 	warnNum := int64(0)
 	critNum := int64(0)
+	missingNum := int64(0)
```
Songmu: How about declaring `var missingFiles []string` and check `len(missingFiles)` instead of `missingNum` for more friendly error messages.

Songmu:
You must check that the value of `--missing` is valid or not before starting to search logs. It is dangerous when typo value is specified like `--missing=WARNNING`. (It is treated as `OK` in your code :scream:)

And why is there no way to specify `--missing=CRITICAL`?

check-log/check-log.go
a-know: I agree. I respected excessively original behavior by mistake. I fixed it d2f7981

check-log/check-log.go
a-know: That's nice. I fixed it fe7f09d

a-know:
Thank you for reviewing.
I added missing option value choices and validation.

check-log/check-log.go
```diff
@@ -78,6 +79,9 @@ func (opts *logOpts) prepare() error {
 			}
 		}
 	}
+	if !(validateMissing(opts.Missing)) {
```
Songmu: needless paren. you can write as follows.

```go
if !validateMissing(opts.Missing) {
```

check-log/check-log.go
```diff
@@ -127,17 +146,32 @@ func run(args []string) *checkers.Checker {
...
+		switch opts.Missing {
+		case "OK":
+		case "WARNING":
+			checkSt = checkers.WARNING
+		case "CRITICAL":
+			checkSt = checkers.CRITICAL
+		default:
+			checkSt = checkers.UNKNOWN
+		}
+		msg += "\n" + fmt.Sprintf("%d files missing as follows.", len(missingFiles))
```
Songmu: English message is strange (?)
How about `The following %d files are missing.`.

check-log/check-log.go
a-know: Thank you!
Fixed it a79cce2

check-log/check-log.go
a-know: Proposed message looks good, I followed it e426d4e

PR #120, opened 2016-10-17
[check-log] [incompatible] Chagne state file naming strategy and location

Songmu:
Name of state file is frequently conflict in current implement. So I change the strategy.

### DIFF
--- a/check-log/check-log.go
+++ b/check-log/check-log.go
@@ -34,6 +34,7 @@ type logOpts struct {
 	patternReg      *regexp.Regexp
 	excludeReg      *regexp.Regexp
 	fileList        []string
+	origArgs        []string
 }
 
 func (opts *logOpts) prepare() error {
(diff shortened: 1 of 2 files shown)


check-log/check-log.go
```diff
@@ -108,8 +110,18 @@ func validateMissing(missing string) bool {
...
+	copy(origArgs, args)
 	opts := &logOpts{}
 	_, err := flags.ParseArgs(opts, args)
+	opts.origArgs = origArgs
+	if opts.StateDir == "" {
+		workdir := os.Getenv("MACKEREL_PLUGIN_WORKDIR")
+		if workdir == "" {
+			workdir = os.TempDir()
+		}
+		opts.StateDir = filepath.Join(workdir, "check-log")
```
astj: [memo]
`StateDir` will be `MkdirAll`ed on `writeBytesToSkip` if necessary, so we don't have to `Mkdir` here

PR #124, opened 2016-11-18
Fix state in check-procs

itchyny:
This pull request fixes the `-s` option of check-procs.

ref: this small bug seems to be introduced in #64


astj:
LGTM!

itchyny:
Thank you!

PR #126, opened 2016-11-24
remove checking Ignore

mattn:
ErrorControl mean how the errors is noticed to users. Not to check the status of process.


daiksy:
LGTM!

PR #129, opened 2016-12-22
add check-windows-eventlog

daiksy:


Makefile
```diff
@@ -35,7 +35,7 @@ cover: devel-deps
 
 build: deps
 	mkdir -p build
-	for i in check-*; do \
+	for i in $(filter-out check-windows-%, $(wildcard check-*)); do \
```
astj: [comment][maybe okay]
Because `make build` accepts `TARGET_OSARCH` param, someone may think `make build TARGET_OSARCH="windows/amd64"` will build executables which run in Windows, but actually it doesn't build `check-windows-%`..

check-windows-eventlog/check_windows_eventlog_test.go
```diff
@@ -0,0 +1,290 @@
+// +build windows
```
astj: [In future]
maybe we should test this repository on Windows CI (AppVeyor?) in future.

astj:
Plugin itself seems good!
In future we should improve build/test procedure especially for Windows, but it's another issue...

LGTM

PR #130, opened 2017-01-02
[check-log] add `file-glob` option

astj:
With this option, users can check logs more flexible, like using `-g /somewhere/*/*/*/*.log` to check against /somewhere/YYYY/MM/DD/N.log.


astj:
Communicated with @Songmu.
We should support globs on `--file`, rather than adding new option `--file-glob`.

PR #133, opened 2017-01-06
[check-log] support glob in --file option

astj:
ref: #131 

It supports both simply glob and zglob (by github.com/mattn/go-zglob).


astj:
I checked these additional tests pass on Windows too. (Sadly `TestGetStateFile` fails so not all tests pass...)

itchyny:
LGTM

astj:
Thanks!

PR #136, opened 2017-01-23
add --source-exclude and --message-exclude, --id

daiksy:
for check-windows-eventlog

### DIFF
--- a/check-windows-eventlog/README.md
+++ b/check-windows-eventlog/README.md
@@ -8,7 +8,7 @@ Checks a windows event log using a regular expression.
 
 ```
 [plugin.checks.event-log]
-command = "/path/to/check-windows-eventlog --log=LOGTYPE --type=EVENTTYPE --source-pattern=REGEXP --source-exclude=REGEXP --message-pattern=REGEXP --message-exclude=REGEXP --warning-over=N --critical-over=N" --fail-first
+command = "/path/to/check-windows-eventlog --log=LOGTYPE --type=EVENTTYPE --source-pattern=REGEXP --source-exclude=REGEXP --message-pattern=REGEXP --message-exclude=REGEXP --event-id-pattern=RANGE --event-id-exclude=RANGE --warning-over=N --critical-over=N" --fail-first
 ```
 
 ### LOGTYPE
@@ -37,11 +37,11 @@ command = "/path/to/check-windows-eventlog --log=LOGTYPE --type=EVENTTYPE --sour
 2. find event which id is 900 or 901.
 
     ```
-    --event-id 900,901
+    --event-id-pattern 900,901
     ```
 
 3. find event which id is between 900 and 1200, but not 1101.
 
     ```
-    --event-id 900-1200,!1101
+    --event-id-pattern 900-1200 --event-id-exclude 1101
     ```
--- a/check-windows-eventlog/README.md
+++ b/check-windows-eventlog/README.md
@@ -37,11 +37,11 @@ command = "/path/to/check-windows-eventlog --log=LOGTYPE --type=EVENTTYPE --sour
 2. find event which id is 900 or 901.
 
     ```
-    --id 900,901
+    --event-id 900,901
     ```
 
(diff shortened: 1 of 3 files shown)


check-windows-eventlog/README.md
```diff
@@ -25,3 +25,23 @@ command = "/path/to/check-windows-eventlog --log=LOGTYPE --type=EVENTTYPE --sour
...
+1. find message matches `foo` but not match `bar`.
+
+    ```
+    --message-pattern foo --message-exclude bar
+    ```
+
+2. find event which id is 900 or 901.
+
+    ```
+    --id 900,901
```
itchyny: IMO, `--event-id` is preferable...

itchyny: Is there any specific reason that `--id` is more suitable for this option?

check-windows-eventlog/check-windows-eventlog.go
```diff
@@ -28,11 +28,25 @@ const (
...
+}
+
 type logOpts struct {
 	Log            string `long:"log" description:"Event Names (comma separated)"`
 	Type           string `long:"type" description:"Event Types (comma separated)"`
 	SourcePattern  string `long:"source-pattern" description:"Event Source (regexp pattern)"`
+	SourceExclude  string `long:"source-exclude" description:"Event Source excluded (regexp pattern)"`
 	MessagePattern string `long:"message-pattern" description:"Message Pattern (regexp pattern)"`
+	MessageExclude string `long:"message-exclude" description:"Message Pattern excluded (regexp pattern)"`
+	EventID        string `long:"event-id" description:"Event IDs (separated by comma)"`
```
astj: Not necessary, but it's better if range specification like `${id1}-${id2}` and excluding by `!${id}` are described.

check-windows-eventlog/check-windows-eventlog.go
```diff
@@ -293,6 +371,30 @@ func (opts *logOpts) searchLog(logName string) (warnNum, critNum int64, errLines
...
+				if !idr.bang {
+					exact = true
+					if idr.lo <= eventID && eventID <= idr.hi {
+						found = true
+					}
+				} else {
+					if idr.lo <= eventID && eventID <= idr.hi {
+						found = false
+					} else if !exact {
+						found = true
```
astj: `!2,!3` against `2` maybe become `found = true`, though `!3,!2` against `2` maybe become `found = false`. Is this behavior accepted? (imo acceptable...)

check-windows-eventlog/README.md
mattn: thanks for your review.
make sense, it looks better.

check-windows-eventlog/check-windows-eventlog.go
mattn: thanks for your review.
hmm, this code is confusible. :(

|No|Seq  |Exact|Found|
|--|-----|-----|-----|
| 1|!2   |false|false|
| 2|!2,!3|false|found|

So `exact` should be true when the first `!` found.

```go
if !idr.bang {
	exact = true
	if idr.lo <= eventID && eventID <= idr.hi {
		found = true
	}
} else {
	if idr.lo <= eventID && eventID <= idr.hi {
		found = false
		exact = true
	} else if !exact {
		found = true
	}
}
```

I'll add test for this edge case. Thanks.

mattn:
At all, I notice this `--event-id` is too difficult to use. And there are more edge case. ex: `2,!3,3`. This will make user confusing. So I suggest to change this like `--source-pattern` and `--source-exclude`.

`--event-id-pattern` and `--event-id-exclude` both allow `1,2` or `3-4`. (don't accept `!`)
And looks IDs following.

1. handle `--event-id-pattern`
2. handle `--event-id-exclude`
    Accepted IDs will be canced

How do you think?

astj:
I saw #137 , and it seems nice to me. How do you think? > @daiksy

daiksy:
@astj 
I think that it is a good thought !

PR #139, opened 2017-01-30
fix matching for 'Audit Success' and 'Audit Failure'

mattn:
It's difficult to specify those values in -type option since toml format
have problem to write value contain spaces.


daiksy:
LGTM!

PR #141, opened 2017-02-16
ensure close temporary file in writeFileAtomically

itchyny:
pointed out by @haya14busa
https://github.com/mackerelio/mackerel-agent/pull/333/files#r101444496

### DIFF
--- a/check-log/check-log.go
+++ b/check-log/check-log.go
@@ -361,8 +361,8 @@ func writeFileAtomically(f string, contents []byte) error {
 		return err
 	}
 	defer os.Remove(tmpf.Name())
-	defer tmpf.Close()
 	_, err = tmpf.Write(contents)
+	tmpf.Close() // should be called before rename
 	if err != nil {
 		return err
 	}
--- a/check-log/check-log.go
+++ b/check-log/check-log.go
@@ -361,8 +361,8 @@ func writeFileAtomically(f string, contents []byte) error {
 		return err
 	}
 	defer os.Remove(tmpf.Name())
-	_, err = tmpf.Write(contents)
 	defer tmpf.Close()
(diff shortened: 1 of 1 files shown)


check-log/check-log.go
```diff
@@ -362,9 +362,9 @@ func writeFileAtomically(f string, contents []byte) error {
 	}
 	defer os.Remove(tmpf.Name())
 	_, err = tmpf.Write(contents)
+	defer tmpf.Close()
```
haya14busa: It's ok to call here but no reason to place this line right after L363? ref: https://github.com/golang/tools/blob/be0fcc31ae2332374e800dfff29b721c585b35df/cmd/goimports/goimports.go#L285-L286

itchyny: OK. I moved the line after `os.Remove`.

itchyny:
I removed `defer` to ensure closing fd before renaming.

check-log/check-log.go
```diff
@@ -361,8 +361,8 @@ func writeFileAtomically(f string, contents []byte) error {
 		return err
 	}
 	defer os.Remove(tmpf.Name())
-	defer tmpf.Close()
 	_, err = tmpf.Write(contents)
+	tmpf.Close() // should be called before rename
```
mattn: Ah, right. The file handle should be closed before Rename. Especially this will be problem on Windows. Sorry. :bowing_man:

haya14busa: :bow:

@itchyny  Could you please write some helpful comment on L363 ?
I saw L363 and then tmpf.Close() can be called by defer too, but L363 doesn't work (or returns error) if rename is done correctly.

Or, I'll create another pull-request if you want.

itchyny:
Thank you for pointing out.

PR #142, opened 2017-02-16
Write the file in same directory

mattn:
Temporary file may not locate at same disk-partition, or same filesytem.
When the path of temporary file doesn't locate at same disk-partition,
os.Rename works like copying file.


itchyny:
Thanks.

PR #144, opened 2017-03-03
Change directory structure convention of each plugin

Songmu:
ref. https://github.com/mackerelio/mackerel-agent-plugins/pull/289

The reason is as follows.

- To increase code reusability
- To suppress the enlargement of the `main` package

This patch is automatically converted by the following script.

https://github.com/mackerelio/go-check-plugins/blob/packagize/tool/packagize.pl


check-windows-eventlog/lib/check-windows-eventlog.go
```diff
@@ -21,7 +21,7 @@ import (
 
 	"github.com/jessevdk/go-flags"
 	"github.com/mackerelio/checkers"
-	"github.com/mackerelio/go-check-plugins/check-windows-eventlog/internal/eventlog"
```
mattn: :bow:

itchyny:
LGTM!

PR #146, opened 2017-03-07
[check-tcp] connect timeout

Songmu:
The `-timeout` option doesn't work when waiting tcp connection, so I fixed.

### DIFF
--- a/check-tcp/check-tcp.go
+++ b/check-tcp/check-tcp.go
@@ -144,7 +144,7 @@ func (opts *tcpOpts) merge(ex exchange) {
 	}
 }
 
-func dial(network, address string, ssl bool, noCheckCertificate bool) (net.Conn, error) {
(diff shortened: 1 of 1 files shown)


check-tcp/check-tcp.go
```diff
@@ -225,23 +228,23 @@ func (opts *tcpOpts) run() *checkers.Checker {
 	return checkers.NewChecker(chkSt, msg)
 }
 
-func write(conn net.Conn, content []byte, timeout float64) error {
-	if timeout > 0 {
-		conn.SetWriteDeadline(time.Now().Add(time.Duration(timeout) * time.Second))
+func write(conn net.Conn, content []byte, timeoutDur time.Duration) error {
```
haya14busa: [nit] I think 'timeout time.Duration' is enough

PR #147, opened 2017-03-07
[check-tcp] [bugfix] fix decimal threshold value handling

Songmu:


check-tcp/check-tcp_test.go
```diff
@@ -235,7 +235,7 @@ func TestUnixDomainSocket(t *testing.T) {
 	testUnexpected()
 
 	testOverWarn := func() {
-		opts, err := parseArgs([]string{"-U", sock, "--send", `PING`, "-E", "-e", "OKOK", "-w", "0.1"})
+		opts, err := parseArgs([]string{"-U", sock, "--send", `PING`, "-E", "-e", "OKOK", "-w", "0.000000001"})
```
mattn: [suggestion] define opts.Critical as time.Duration. you can specify this value like `-w 3s`.

Songmu: I know that, but there are some reasons for not doing so.

- It always requires postfix
  - I want to specify only the seconds
- check-tcp partially follows interface of Nagios' check_tcp. The `-w` option is also follows.
- It is breaking change

PR #156, opened 2017-03-23
check lower-case driver letters

mattn:
When lower-case filename is specified with `-file`, check-log try to make directory `%APPDATA%\temp\check-log\C:`. (`:` is invalid)


Songmu:
Thanks!

PR #157, opened 2017-03-26
use wmi query instead of running wmic command

mattn:
execute wmic command taks bits heavy CPU/Memory loads.


daiksy:
LGTM!

mattn:
:golf: Thank you! :golfing_man:

PR #171, opened 2017-05-16
[check-ssh] fix the problem that check-ssh cannot invoke SSH connection

astj:
ref: https://github.com/golang/go/issues/19767


Songmu:
LGTM

astj:
Thanks!

PR #174, opened 2017-05-22
[check-log]  When specified multiple pattern, perform search that satisfies all conditions

a-know:
Current specification of `check-log` , it is difficult to satisfy `AND` as the log monitoring condition.

This p-r is to make it possible to perform log monitoring that satisfies all the patterns when multiple `--pattern` are specified.

### DIFF
--- a/check-log/lib/check-log.go
+++ b/check-log/lib/check-log.go
@@ -23,7 +23,7 @@ import (
 
 type logOpts struct {
 	LogFile             string   `short:"f" long:"file" value-name:"FILE" description:"Path to log file"`
-	Pattern             []string `short:"p" long:"pattern" required:"true" value-name:"PAT" description:"Pattern to search for. If specified multiple, they are treated as AND condition"`
+	Pattern             []string `short:"p" long:"pattern" required:"true" value-name:"PAT" description:"Pattern to search for. If specified multiple, they will be treated together with the AND operator"`
 	Exclude             string   `short:"E" long:"exclude" value-name:"PAT" description:"Pattern to exclude from matching"`
 	WarnOver            int64    `short:"w" long:"warning-over" description:"Trigger a warning if matched lines is over a number"`
 	CritOver            int64    `short:"c" long:"critical-over" description:"Trigger a critical if matched lines is over a number"`
--- a/check-log/lib/check-log.go
+++ b/check-log/lib/check-log.go
@@ -23,7 +23,7 @@ import (
 
 type logOpts struct {
 	LogFile             string   `short:"f" long:"file" value-name:"FILE" description:"Path to log file"`
-	Pattern             []string `short:"p" long:"pattern" required:"true" value-name:"PAT" description:"Pattern to search for"`
+	Pattern             []string `short:"p" long:"pattern" required:"true" value-name:"PAT" description:"Pattern to search for. If specified multiple, they are treated as AND condition"`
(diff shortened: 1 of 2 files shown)


stefafafan:
The code seems correct, just not sure if, the points below are okay.
- specifying multiple `--pattern` 
- show error when warning/critical is specified for multiple

Also, if we are going to allow multiple `--pattern` , it may be nice to also write that in the README of checklog.

check-log/lib/check-log.go
```diff
@@ -309,11 +318,16 @@ func (opts *logOpts) searchReader(rdr io.Reader) (warnNum, critNum, readBytes in
...
-	matched := len(matches) > 0 && (eReg == nil || !eReg.MatchString(line))
+	matched := true
+	var matches []string
+	for _, pReg := range opts.patternReg {
+		if matched {
+			eReg := opts.excludeReg
+
+			matches = pReg.FindStringSubmatch(line)
+			matched = len(matches) > 0 && (eReg == nil || !eReg.MatchString(line))
+		}
```
mattn:
```go
for _, pReg := range opts.patternReg {
    eReg := opts.excludeReg

    matches = pReg.FindStringSubmatch(line)
    if len(matches) > 0 && (eReg == nil || !eReg.MatchString(line)) {
        return true, matches
    }
    return false, nil
}
```

a-know: Thanks for review!
In this case, it is aimed to detect matching all patterns in `opts.patternReg`, so the correct code is below?


```go
for _, pReg := range opts.patternReg {
    eReg := opts.excludeReg

    matches = pReg.FindStringSubmatch(line)
    if len(matches) == 0 || (eReg != nil && eReg.MatchString(line)) {
        return false, nil
    }
}
return true, matches
```

mattn: Seems good to me :+1:

check-log/lib/check-log.go
```diff
@@ -22,21 +22,21 @@ import (
...
-	ReturnContent       bool    `short:"r" long:"return" description:"Return matched line"`
-	FilePattern         string  `short:"F" long:"file-pattern" value-name:"FILE" description:"Check a pattern of files, instead of one file"`
-	CaseInsensitive     bool    `short:"i" long:"icase" description:"Run a case insensitive match"`
-	StateDir            string  `short:"s" long:"state-dir" value-name:"DIR" description:"Dir to keep state files under"`
-	NoState             bool    `long:"no-state" description:"Don't use state file and read whole logs"`
-	Encoding            string  `long:"encoding" description:"Encoding of log file"`
-	Missing             string  `long:"missing" default:"UNKNOWN" value-name:"(CRITICAL|WARNING|OK|UNKNOWN)" description:"Exit status when log files missing"`
-	patternReg          *regexp.Regexp
+	LogFile             string   `short:"f" long:"file" value-name:"FILE" description:"Path to log file"`
+	Pattern             []string `short:"p" long:"pattern" required:"true" value-name:"PAT" description:"Pattern to search for. If specified multiple, they are treated as AND condition"`
```
stefafafan: > If specified multiple, they are treated as AND condition

[nit][english]
How about something like,
`If specified multiple, they will be treated together with the AND operator.`

a-know: Thanks!
I fixed it by ddf197b .

itchyny:
LGTM

PR #175, opened 2017-05-28
v2 packages (rpm and deb)

Songmu:
Add build scripts and settings for v2 packages.

- rpm-v2
  - mackerel-check-plugins-x.y.z-1.el7.centos.x86_64.rpm
  - CentOS7 or later
- deb-v2
  - mackerel-check-plugins_x.y.z-1.v2_amd64.deb
  - ubuntu 16.04, debian 8 or later

Regarding the old packages, I adjusted the format not to break the build in future.

## Differences from v1

- provides for 64bit architecture only
- package contains only one binary file (mackerel-plugin) and symlinks


astj:
LGTM

PR #178, opened 2017-06-27
add check-disk

edangelion:


check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,121 @@
...
+			units = b
+		case *opts.Units == "kB":
+			units = kb
+		case *opts.Units == "MB":
+			units = mb
+		case *opts.Units == "GB":
+			units = gb
+		case *opts.Units == "TB":
+			units = tb
+		default:
```
itchyny: Reject invalid argument here. (`check-disk -u tb` does not work...)

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,121 @@
...
+	All   uint64 `json:"all"`
+	Used  uint64 `json:"used"`
+	Free  uint64 `json:"free"`
+	Avail uint64 `json:"available"`
+}
+
+var opts struct {
+	Warning      *float64 `short:"W" long:"warning" value-name:"N" description:"Exit with WARNING status if less than N units of disk are free"`
+	Critical     *float64 `short:"C" long:"critical" value-name:"N" description:"Exit with CRITICAL status if less than N units of disk are free"`
+	WarningRate  *float64 `short:"w" long:"warning-rate" value-name:"N" description:"Exit with WARNING status if less than N % of disk are free"`
```
itchyny: I think the behavior of the original command is nice because the user do not confused by which to use -w or -W. Accept as `*string` and parse it.

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,121 @@
...
+	if opts.WarningRate != nil && *opts.WarningRate > freeRate {
+		checkSt = checkers.WARNING
+	}
+	if opts.CriticalRate != nil && *opts.CriticalRate > freeRate {
+		checkSt = checkers.CRITICAL
+	}
+
+	u := "MB"
+	if opts.Units != nil {
+		u = *opts.Units
```
itchyny: `check-disk -u foo` should not work IMO...

edangelion:
Thanks, I changed codes.

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,148 @@
...
+		v, err := strconv.Atoi(strings.TrimRight(val, "%"))
+		if err != nil {
+			return checkers.UNKNOWN, err
+		}
+
+		if float64(v) > freePct {
+			checkSt = checkers.WARNING
+		}
+	} else {
+		v, err := strconv.Atoi(val)
```
astj: [nit]
Since `v` will be converted into float64 later, is it possible to accept float64 `v` value and use `ParseFloat` here?

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,148 @@
...
+	disk := &diskStatus{}
+	disk.All = fs.Blocks * uint64(fs.Bsize)
+	disk.Free = fs.Bfree * uint64(fs.Bsize)
+	disk.Used = disk.All - disk.Free
+	disk.Avail = fs.Bavail * uint64(fs.Bsize)
+
+	return disk, nil
+}
+
+func checkStatus(val string, units float64, disk *diskStatus) (checkers.Status, error) {
```
astj: `val` is not obvious what it is, how about using like `threshold` ?

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,148 @@
...
+			checkSt = checkers.WARNING
+		}
+	} else {
+		v, err := strconv.Atoi(val)
+		if err != nil {
+			return checkers.UNKNOWN, err
+		}
+
+		if float64(v) > avail {
+			checkSt = checkers.WARNING
```
astj: checkStatus is also used with `opts.Critical`, but this implementation never returns `checkers.CRITICAL`.
So you should do one of following:
- receive checker.Status param to use on exceeding threshold
- return some signature to tell exceeding, not checkers.Status itself

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,148 @@
...
+	"strconv"
+	"strings"
+	"syscall"
+
+	"github.com/jessevdk/go-flags"
+	"github.com/mackerelio/checkers"
+)
+
+type diskStatus struct {
+	All   uint64 `json:"all"`
```
astj: json tags seems unused?

check-disk/lib/check-disk.go
edangelion: Oops...Thank you.

edangelion:
Thank you, I have fixed codes.

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,148 @@
...
+	if opts.Units != nil {
+		us = *opts.Units
+	}
+
+	all := float64(disk.All) / float64(units)
+	used := float64(disk.Used) / float64(units)
+	free := float64(disk.Free) / float64(units)
+	avail := float64(disk.Avail) / float64(units)
+	freePct := (float64(disk.Avail) * float64(100)) / float64(disk.All)
+	msg := fmt.Sprintf("All: %.2f %v, Used: %.2f %v, Free: %.2f %v, Available: %.2f %v, Free percentage: %.2f\n", all, us, used, us, free, us, avail, us, freePct)
```
astj: [nit]
Is it useful if the message includes threshold, especially when status is not ok?
(like: `Free Percentage 29 < 30` ?)

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,148 @@
...
+	ckr.Exit()
+}
+
+func run(args []string) *checkers.Checker {
+	_, err := flags.ParseArgs(&opts, args)
+	if err != nil {
+		os.Exit(1)
+	}
+
+	path := "/"
```
astj: Hmm…
nagios's check_disk checks against all partitions by default, so this behavior may be confusing.

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,212 @@
...
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/jessevdk/go-flags"
+	"github.com/mackerelio/checkers"
+	gpud "github.com/shirou/gopsutil/disk"
+)
+
+var opts struct {
```
astj: [Suggestion]
Original check_disk has so many options. I don't think check-disk must implement all of them, but some of them are not hard to implement, so please consider it. (Or you can do it in another p-r.)
- inode monitoring threshold `-W`/`-K`
- fstype filtering `-X`/`-N`
- Explicit check against all partitions `-A`

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,212 @@
...
+	Warning  *string `short:"w" long:"warning" value-name:"N, N%" description:"Exit with WARNING status if less than N units or N% of disk are free"`
+	Critical *string `short:"c" long:"critical" value-name:"N, N%" description:"Exit with CRITICAL status if less than N units or N% of disk are free"`
+	Path     *string `short:"p" long:"path" value-name:"PATH" description:"Mount point or block device as emitted by the mount(8) command"`
+	Exclude  *string `short:"x" long:"exclude_device" value-name:"EXCLUDE PATH" description:"Ignore device (only works if -p unspecified)"`
```
astj: It's more user-friendly if plugin warns (or dies?) when both `-p` and `-x` are specified.

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,212 @@
...
+
+	"github.com/jessevdk/go-flags"
+	"github.com/mackerelio/checkers"
+	gpud "github.com/shirou/gopsutil/disk"
+)
+
+var opts struct {
+	Warning  *string `short:"w" long:"warning" value-name:"N, N%" description:"Exit with WARNING status if less than N units or N% of disk are free"`
+	Critical *string `short:"c" long:"critical" value-name:"N, N%" description:"Exit with CRITICAL status if less than N units or N% of disk are free"`
+	Path     *string `short:"p" long:"path" value-name:"PATH" description:"Mount point or block device as emitted by the mount(8) command"`
```
astj: check_disk has `-R`/`-r` options to select paths by regexp pattern.  Do you have any idea how to handle regexp pattern? (Simply implement those options, accept patterns in `-p`, etc...)

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,212 @@
...
+	Exclude  *string `short:"x" long:"exclude_device" value-name:"EXCLUDE PATH" description:"Ignore device (only works if -p unspecified)"`
+	Units    *string `short:"u" long:"units" value-name:"STRING" description:"Choose bytes, kB, MB, GB, TB (default: MB)"`
```
astj: Should we support options like `--megabytes`? (check_disk has SO many options!!)

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,212 @@
...
+
+	if opts.Critical != nil {
+		for _, disk := range disks {
+			checkSt, err = checkStatus(checkSt, *opts.Critical, u.Size, disk, checkers.CRITICAL)
+			if err != nil {
+				return checkers.Unknown(fmt.Sprintf("Faild to check disk status: %s", err))
+			}
+
+			if checkSt == checkers.CRITICAL {
+				break
```
astj: [nit]
If you can check critical first,  you can skip checking warnings when critical error happens

check-disk/lib/check-disk.go
```diff
@@ -0,0 +1,302 @@
...
+
+	var disks []*gpud.UsageStat
+
+	for _, partition := range partitions {
+		disk, err := gpud.Usage(partition.Mountpoint)
+		if err != nil {
+			return checkers.Unknown(fmt.Sprintf("Faild to fetch disk usage: %s", err))
+		}
+
+		disks = append(disks, disk)
```
astj: Original `check_disk` skips partitions which has 0 blocks (= `disk.Total == 0`).

edangelion:
I changed codes. Thank you.

PR #183, opened 2017-07-21
Status map

### LINKED ISSUE #181: check_http.go - Override statuses?
Hi,

I'm currently working on a web server with IP blacklist enabled.
If a blacklisted client accesses it, it should return 403.

I would like to make sure such a control is indeed working-- in order to achieve that, I want `check_http` to return `OK` when the web server returns 403 under a certain condition (Particularly #182 ) and `CRITICAL` when it returns anything else, including 200 OK.

I'm concerned that it might be a long way to go, but I'd be very happy to use `check_http` in this way.
Thanks.

mattn:
Possible patterns:
```
  -s 404=ok
  -s 300-404=critical
```
Fixes #181

### DIFF
--- a/check-http/README.md
+++ b/check-http/README.md
@@ -9,4 +9,10 @@ Check http.
 check-http -u http://example.com
 ```
 
+To override status
+```shell
+check-http -s 404=ok -u http://example.com
+check-http -s 200-404=ok -u http://example.com
+```
+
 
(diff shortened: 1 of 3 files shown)


chris-x86-64:
I'm amazed how quick you got this! Way to go!

syou6162:
Thank you for sending your pull request. LGTM!

PR #184, opened 2017-07-24
add -i flag to specify source IP

### LINKED ISSUE #182: check_http.go - Source IP address?
This issue relates to https://github.com/mackerelio/go-check-plugins/issues/181 ...

In conjunction with #181, I want `check_http` to be able to use a specific source IP address or network interface, similar to curl's `--interface` option.

With curl, you can do something like this:
```
curl --interface 127.0.0.127 http://127.0.0.1
```

Can I achieve the same using `check_http` ?

Thanks.

mattn:
Fixes #182

### DIFF
--- a/check-http/lib/check_http.go
+++ b/check-http/lib/check_http.go
@@ -108,8 +108,12 @@ func run(args []string) *checkers.Checker {
 		Proxy: http.ProxyFromEnvironment,
 	}
 	if opts.SourceIP != "" {
+		ip := net.ParseIP(opts.SourceIP)
+		if ip == nil {
+			return checkers.Unknown(fmt.Sprintf("Invalid source IP address: %v", opts.SourceIP))
+		}
 		tr.Dial = (&net.Dialer{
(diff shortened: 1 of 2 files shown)


check-http/lib/check_http.go
```diff
@@ -105,6 +107,14 @@ func run(args []string) *checkers.Checker {
 		},
 		Proxy: http.ProxyFromEnvironment,
 	}
+	if opts.SourceIP != "" {
+		tr.Dial = (&net.Dialer{
+			LocalAddr: &net.TCPAddr{IP: net.ParseIP(opts.SourceIP)},
```
syou6162: `net.ParseIP` returns nil if the given string is not a valid representation of an IP address. In such cases, returning `Unknown` would be plausible in my opinion.

mattn: Thank you. Fixes.

syou6162:
Thank you for sending great pull request. LGTM!

PR #190, opened 2017-08-18
[check-log] Add check-first option and skip the log file on the first run on default

edangelion:


mattn:
IMO, this should be default behavior. And flag name should be failfirst or something.

mattn:
As far as I think, most of users expect -skipfirst=true in default. For example, the case that restart system after errors something raised. user doesn't want to see duplicated alarms, I think.

itchyny:
Thanks, we'll reconsider the spec.

itchyny:
LGTM!

mattn:
LGTM :+1:

PR #200, opened 2017-10-13
fix check-disk options: -x, -X, -p, -N

plaster:
- `-p` / `-N`
  - fixed output inflation when a value specified twice (ex. -N ext4 -N ext4)
- `-x`
  - now accpets twice
- `-X`
  - fixed exclusion failure when specified twice


astj:
Thank you for contributing. LGTM.

plaster:
Thank you for merging!

PR #202, opened 2017-11-02
Skip IPv6 tests in Travis

astj:
TestHTTPIPv6 in check-tcp tests start failing recently: https://travis-ci.org/mackerelio/go-check-plugins/builds
From the error message, maybe ipv6 networking does not work in Travis container. (Even if so, I don't know why it has been working...)

So I'd like to skip this test for now.


itchyny:
LGTM

PR #203, opened 2017-11-28
[redis] check replication role

y-kuno:


check-redis/lib/check-redis.go
```diff
@@ -126,11 +126,11 @@ func checkReachable(args []string) *checkers.Checker {
 	}
 	defer c.Close()
 
-	if version, ok := (*info)["redis_version"]; !ok {
+	version, ok := (*info)["redis_version"];
```
syou6162: Original one (`if version, ok := (*info)["redis_version"]; !ok {`) would be betteer.

ref: https://golang.org/doc/effective_go.html#if

check-redis/lib/check-redis.go
```diff
@@ -151,6 +149,12 @@ func checkSlave(args []string) *checkers.Checker {
 	}
 	defer c.Close()
 
+	if role, ok := (*info)["role"]; !ok {
+		return checkers.Unknown("couldn't get role")
+	} else if role == "master" {
+		return checkers.Ok("redis is not slave")
```
syou6162: When we use `check-redis slave ...`, we want to check the slave status. This condition matches the case when we wrongly specify the master node for `--host` option.

So `return checkers.Unknown("redis is not slave")` is better because we can notice that we wrongly specify `--host` option.

syou6162:
@y-kuno Thank you for sending pull request! I wrote some comments, please take a look.

syou6162:
Sorry but closing the pull request for now.

PR #204, opened 2017-12-06
[check-procs] Error handling is required

mechairoi:


itchyny:
LGTM.
It seems that this p/r changes the exit status on invalid arguments or etc, but the errors are actually fatal error so returning UNKNOWN statuses look good to me.

PR #205, opened 2017-12-18
[check-mysql] add readonly subcommand

ichirin2501:
check-mysqlでreadonlyの監視機能を追加してみました

### DIFF
--- a/check-mysql/lib/readonly.go
+++ b/check-mysql/lib/readonly.go
@@ -1,6 +1,7 @@
 package checkmysql
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/jessevdk/go-flags"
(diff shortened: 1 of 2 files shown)


check-mysql/lib/readonly.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"os"
+
+	"github.com/jessevdk/go-flags"
+	"github.com/mackerelio/checkers"
+)
+
+type readOnlyOpts struct {
+	mysqlSetting
```
astj: It's better to accept options to specify which status to expect, IS readonly or IS NOT readonly.

ichirin2501: I see now.
I added a flag parameters and committed it. https://github.com/mackerelio/go-check-plugins/pull/205/commits/b25ceaedecc4a5f01102186752bc406c89a3ea5c

check-mysql/lib/readonly.go
```diff
@@ -0,0 +1,47 @@
...
+	NoReadOnly bool `short:"n" long:"no-readonly" description:"expect that it is not read_only with the MySQL parameter"`
+}
+
+func isReadOnly(v string) bool {
```
itchyny: This function looks weird to me... maybe inlining is better for this...

itchyny:
How about this? `check-mysql readonly on`, `check-mysql readonly off`

check-mysql/lib/readonly.go
ichirin2501: ON / OFF of character string is dependent on MySQL side. On the program side I had the reason that I wanted to convert it once and treat it as True / False.

There is no strong reason so I will fix it. (inlining)

check-mysql/lib/readonly.go
```diff
@@ -0,0 +1,47 @@
...
+	rows, res, err := db.Query("SHOW GLOBAL VARIABLES LIKE 'read_only'")
+	if err != nil {
+		return checkers.Unknown("couldn't execute query")
+	}
+
+	idxReadOnly := res.Map("Value")
+	readOnlyStatus := rows[0].Str(idxReadOnly)
+
+	if isReadOnly(readOnlyStatus) == opts.NoReadOnly {
+		return checkers.Critical("the expected value of read_only is different")
```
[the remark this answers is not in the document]
ichirin2501: ok !!

ichirin2501:
> How about this? `check-mysql readonly on`, `check-mysql readonly off`

Wow, that is simple and good. I would like to do so with arguments instead of options.

ichirin2501:
fix commit! https://github.com/mackerelio/go-check-plugins/pull/205/commits/e31287fd4c2f9bf7fe4ca3133b0a6f6c9f924d41

astj:
Thank you!

PR #210, opened 2018-01-23
[check-http] Expose checkhttp.Run due for using check-http as a library

aereal:
[go-magi][], yet another external monitoring tool, currently execute check-http command using os/exec package.

Generally speaking, we must pay some costs while calling fork or exec syscalls, and also have to abandon cross-platform features while touching `ProcessState`, even writing in Go.

[go-magi]: https://github.com/aereal/go-magi


Songmu:
:+1:

PR #212, opened 2018-01-23
[check-tcp] add -W option

dozen:
I Added -W option in order to suppress CRITICAL when error occurred.

-W sets the exit status at error occurrence to WARNING.

In the case of request succeeded, evaluation result of -c option eval takes priority.


dozen:
I erased the branch by mistake.
I will repoen 🙇

Songmu:
Good pull request!

Could you rename `--err-warning` to `--error-warning`? Because it is a policy not to use abbreviations in the long option on check plugins.

dozen:
I agree. Fixed.

Songmu:
Thank you!

PR #215, opened 2018-01-25
[check-uptime] add detect rewind option

dozen:
I want to detect reboot.

So, I added `-r, --warning-rewind` and `-R, --critical-rewind` options.

Alerts when the value is smaller than the previous operation time.


dozen:
I solved the problem without changing this plugin. So, I will close.

PR #222, opened 2018-02-13
[check-log]improve skip bytes when file size is zero

hayajo:
- If the file size is zero, the following error will be ignored.


itchyny:
LGTM

mattn:
reducing unnecessary syscall?
```
if !opts.NoState && !opts.CheckFirst {
	if _, err = os.Stat(stateFile); os.IsNotExist(err) {
		skipBytes = stat.Size()
	}
}
```

hayajo:
LGTM
Thank you! @mattn

PR #224, opened 2018-03-09
[check-http] add host header option

taku-k:
* For those who want to use host header option.
* check_http plugin has the same option.

### DIFF
--- a/check-http/lib/check_http.go
+++ b/check-http/lib/check_http.go
@@ -1,11 +1,13 @@
 package checkhttp
 
 import (
+	"bufio"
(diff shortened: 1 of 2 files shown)


mattn:
I prefer to add -H option as slice for specifying multiple headers like below.

```
-H "Host: foobar" -H "Authorization: Basic XXXXXXXXXXXXXXXXXXXXX"
```

taku-k:
That's so flexible. I try to implement what you proposed. Thank you.

hayajo:
Thank you!

PR #225, opened 2018-03-12
[check-http] add expected string option for check-http

taku-k:
* This option makes sure the content has the expected string.

### DIFF
--- a/check-http/lib/check_http.go
+++ b/check-http/lib/check_http.go
@@ -1,6 +1,7 @@
 package checkhttp
 
 import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


mattn:
I would appreciate it if some opinions, I prefer this option works as RegExp pattern. (Please do not change)

hayajo:
I also like to work as a RegExp pattern.
For example, what about changing to the `--pattern` flag and acting as a RegExp pattern?

hayajo:
LGTM!
Thank you!!

PR #227, opened 2018-03-20
add check-ldap

taku-k:


yuuki:
Here is the original implementation https://www.monitoring-plugins.org/doc/man/check_ldap.html .

hayajo:
LGTM!
Thank you!!

PR #231, opened 2018-04-18
Fix duplicated output of usage with --help argument

itchyny:
This pull request fixes the problem `check-procs --help` and `check-cert-file --help` outputs the command usage twice.

```sh
 % for f in check-*; do (cd $f >/dev/null && test $(go run main.go --help 2>&1 | grep 'Application Options' | wc -l) -gt 1 && echo $f); done
check-cert-file/
check-procs/
```


itchyny:
Thanks!

PR #233, opened 2018-04-27
[check-windows-eventlog] fix README - Some of the listed EVENTTYPEs can not be detected as alerts

a-know:
The following EVENTTYPE can not be detected as an alert.

- Success
- Audit Success
- Information

I think that it is better to mention this in the README.


itchyny:
LGTM

a-know:
Thanks!!

PR #234, opened 2018-05-03
[check-log] Add option to suppress pattern display

k-hal:
Added pattern display suppression option because it is hard to see when the target pattern becomes longer.
Assuming use with `--return` option.


itchyny:
LGTM

PR #236, opened 2018-05-25
[check-mysql] add unix socket option

sugy:
See also.
- https://github.com/mackerelio/mackerel-agent-plugins/blob/master/mackerel-plugin-mysql/lib/mysql.go#L325-L329
- https://github.com/mackerelio/mackerel-agent-plugins/blob/master/mackerel-plugin-mysql/lib/mysql.go#L1030-L1035


hayajo:
LGTM
Thank You!

PR #237, opened 2018-06-05
Add feature use ntp server

netmarkjp:
In our usecase, compare localtime to ntp server is better (just like nagios plugins).

I think compare localtime to ntp server is more direct(and have some benefit) way than check via time-sync daemon.


hayajo:
LGTM
Thank You!

PR #239, opened 2018-06-12
[check-http] Set Host header via req.Host

nonylene:
net/http uses `req.Host` for Host header, not `req.Headers`.
`req.Host` is set to `req.URL.Host` by default, so `Host` in `req.Headers` is ignored.

In this commit, set `req.Host` from parsed header if specified.

> For client requests, certain headers such as Content-Length
> and Connection are automatically written when needed and
> values in Header may be ignored. See the documentation
> for the Request.Write method.
>
> https://golang.org/pkg/net/http/#Request


susisu:
Nice catch! Thank you.

PR #241, opened 2018-07-06
modify message check-windows-eventlog

daiksy:
modified message because many asking about the meaning of error messages

### DIFF
--- a/check-windows-eventlog/lib/check-windows-eventlog.go
+++ b/check-windows-eventlog/lib/check-windows-eventlog.go
@@ -471,7 +471,7 @@ loop_events:
 
 		if opts.ReturnContent {
 			if message == "" {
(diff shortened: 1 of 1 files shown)


check-windows-eventlog/lib/check-windows-eventlog.go
```diff
@@ -471,7 +471,7 @@ loop_events:
 
 		if opts.ReturnContent {
 			if message == "" {
-				message = "[mackerel-agent] No message resource found. Please make sure the event log occured on the server."
+				message = "Because the message resource could not be found, the event log message could not be obtained by mackerel-agent. Please access the target server and check the event log directly."
```
astj: Since users can execute check plugins without mackerel-agent, error messages from those plugins should not refer mackerel-agent (unless the error related mackerel-agent itself).
How about omitting `by mackerel-agent` ?

PR #250, opened 2018-09-06
[check-log] Trace an old file after logrotation with the inode number 

yuuki:
# Problem 

`check-log` skips to check a renamed old file after logrotation.

# Resolution

1. saves inode number of the target logfile into a state file 
1. logrotate (ex. mv `/var/log/access.log` => `/var/log/access.log.1` and create `/var/log/access.log`)
1. find a file with same inode number of the logfile checked the last time in the directory of the target logfile.
1. regards the found file as rotated file and also checks the file. 

# Notes

- `copytruncate` is not supported.
- windows is not supported.


check-log/lib/check-log.go
```diff
@@ -230,13 +231,26 @@ func (opts *logOpts) searchLog(logFile string) (int64, int64, string, error) {
 		skipBytes = s
 	}
 
-	f, err := os.Open(logFile)
-	if err != nil {
-		return 0, 0, "", err
+	var r io.ReadSeeker
+	if opts.NoState || opts.CheckFirst {
```
astj: We should trace logrotate when CheckFirst option is specified, because current behavior is:

### When CheckFirst is disabled
- On first execution, just record current file size to the state file
    - At this time check-log does not check the contents of the file
- On second execution (and afterwards):
    1. Seek to previously recorded position (or go back to head if current file size is smaller than the position)
    2. Check the contents between sought position and EOF
    3. Record EOF position (aka file length) to state file

### When CheckFirst is enabled
- On first execution:
    1. Check the contents in all of the log file
    2. Record ROF position (aka file length) to state file
- On second execution (and afterwards):
    1. Seek to previously recorded position (or go back to head if current file size is smaller than the position)
    2. Check the contents between sought position and EOF
    3. Record EOF position (aka file length) to state file


As below, CheckFirst-enabled check-log behaves differently only at first execution, so it should trace logrotate same as Check-First disabled check-log. (Of course, it's apparently unnecessary when NoState enabled.)

yuuki: Thanks for your review.
I misunderstood the CheckFirst behavior, so I will accept your comment.

yuuki:
I solved the problem that the state file management was duplicated by check-log and Songmu/postailer.
 
1. Jsonize check-logs's status file https://github.com/mackerelio/go-check-plugins/pull/252
1. Include inode number into the JSON status file
1. add tracing inode logic to check-log without postailer

yuuki:
@astj Review again please!

check-log/lib/check-log.go
```diff
@@ -269,14 +285,30 @@ func (opts *logOpts) searchLog(logFile string) (int64, int64, string, error) {
...
+			oldErrLines            string
+		)
+		// ignore readBytes under the premise that the old file will never be updated.
+		oldWarnNum, oldCritNum, _, oldErrLines, err := opts.searchReader(oldr)
+		if err != nil {
+			return oldWarnNum, critNum, errLines, err
+		}
+		warnNum += oldWarnNum
+		critNum += oldCritNum
+		errLines += oldErrLines
```
astj: When `--return` option specified, matched lines will be returned.
```
$ go run main.go --check-first -f README.md --pattern ISO --warning-over=1 --return
LOG CRITICAL: 14 warnings, 14 criticals for pattern /ISO/.
[README.md]
* ISO-2022-JP
* ISO-8859-1
* ISO-8859-2
* ISO-8859-3
* ISO-8859-4
* ISO-8859-5
* ISO-8859-6
* ISO-8859-7
* ISO-8859-8
* ISO-8859-10
* ISO-8859-13
* ISO-8859-14
* ISO-8859-15
* ISO-8859-16

exit status 2
```
With this implementation, matched lines in old file and those in new file will be displayed in same manner and users cannot distinguish. Do you have any idea to solve this problem?
(Though I think this p-r's behavior is acceptable)

yuuki: That's a good point.
I can add a prefix like `[old]` and `[path/to/file]`to displayed lines, but check-log will lost backward compatibility if a user's program parses check-logs's output.

yuuki:
@astj I fixed https://github.com/mackerelio/go-check-plugins/pull/250#discussion_r216944207.

PR #252, opened 2018-09-07
[check-log] Jsonize status file

yuuki:
It is for future extensions such as https://github.com/mackerelio/go-check-plugins/pull/250. 
This PR supports backward compatibility for old style of status file.

### DIFF
--- a/check-log/lib/check-log.go
+++ b/check-log/lib/check-log.go
@@ -384,7 +384,7 @@ func getBytesToSkip(f string) (int64, error) {
 		return state.SkipBytes, err
 	}
 	// Fallback to read old style status file
-	// for backward compability.
+	// for backward compatibility.
 	// Once saved as new style file, the following will be unreachable.
 	oldf := strings.TrimSuffix(f, ".json")
 	return getBytesToSkipOld(oldf)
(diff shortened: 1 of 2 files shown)


check-log/lib/check-log.go
```diff
@@ -346,20 +347,50 @@ func (opts *logOpts) match(line string) (bool, []string) {
...
-			"%s-%x",
+			"%s-%x.json",
 			stateRe.ReplaceAllString(f, `$1`+string(filepath.Separator)),
 			md5.Sum([]byte(strings.Join(args, " "))),
 		),
 	)
 }
 
 func getBytesToSkip(f string) (int64, error) {
+	if _, err := os.Stat(f); err == nil {
```
astj: loadState itself also checks file's existence, so when json stat file exists `os.Stat` will be executed twice.
Do you have any idea to solve this?
(an idea: Make loadState return `(nil, nil)` when file does not exist, and handle nil here)

yuuki: Thanks. 
I solved at https://github.com/mackerelio/go-check-plugins/pull/252/commits/1724b13f4a5280a649f3574179ba910674ab183f.

yuuki:
@astj Do you have any other comment?

PR #261, opened 2018-10-18
[check-redis] add replication subcommand

yuuki:
add replication subcommand insteam of slave
because it seems that the behavior that slave command returns OK if role is master is not intuitive in the PR(https://github.com/mackerelio/go-check-plugins/pull/259).

```
[y_uuki@redis01 ~]$ redis-cli info | grep role
role:master
[y_uuki@redis01 ~]$ ./check-redis replication
Redis Replication UNKNOWN: couldn't get master_link_status
[y_uuki@redis01 ~]$ ./check-redis replication --skip-master
Redis Replication OK: role is master
```

```
[y_uuki@redis002 ~]$ redis-cli info | grep role
role:slave
[y_uuki@redis002 ~]$ ./check-redis replication
Redis Replication OK: master_link_status: up
```

### DIFF
--- a/check-redis/lib/check-redis.go
+++ b/check-redis/lib/check-redis.go
@@ -21,8 +21,9 @@ type redisSetting struct {
 }
 
 var commands = map[string](func([]string) *checkers.Checker){
-	"reachable": checkReachable,
-	"slave":     checkSlave,
+	"reachable":   checkReachable,
+	"replication": checkReplication,
+	"slave":       checkSlave,
 }
 
 func separateSub(argv []string) (string, []string) {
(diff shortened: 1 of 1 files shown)


check-redis/lib/check-redis.go
```diff
@@ -135,10 +136,59 @@ func checkReachable(args []string) *checkers.Checker {
...
+	}
+
+	c, info, err := connectRedisGetInfo(opts.redisSetting)
+	if err != nil {
+		return checkers.Unknown(err.Error())
+	}
+	defer c.Close()
+
+	if role, ok := (*info)["role"]; ok {
+		if role != "slave" && opts.SkipMaster {
```
astj: When role is "sentinel" and `--skip-master` is specified, this implementation reports as ok. but imo in that case it should return unknown?
https://redis.io/commands/role

astj: Oh sorry, `role` in `info` cannot be `sentinel`? If so, it's no problem.
https://redis.io/commands/INFO

PR #265, opened 2018-12-14
[check-disk] skip checks if there are no inodes

susisu:
On Windows, there are no "inodes" and checks on them should be skipped.
However, the current version always raises wranings / errors if `-w N%` / `-c N%` is specified.
This PR fixes it by skipping checks if total inodes is zero (`gopsutil/disk` returns `InodesTotal = 0` on Windows https://github.com/shirou/gopsutil/blob/master/disk/disk_windows.go).

The change may affect other (non-Windows) environments, but I think it is OK because talking about the "percentage" of 0/0 is nonsense...

### DIFF
--- a/check-disk/lib/check-disk.go
+++ b/check-disk/lib/check-disk.go
@@ -48,6 +48,7 @@ func checkStatus(current checkers.Status, threshold string, units float64, disk 
 		freePct := float64(100) - disk.UsedPercent
 		inodesFreePct := float64(100) - disk.InodesUsedPercent
 
+		// Skip checking if disk.InodesTotal == 0 since inodesFreePct is meaningless.
 		if chkInode && (disk.InodesTotal != 0 && thresholdPct > inodesFreePct) {
 			current = status
 		}
--- a/check-disk/lib/check-disk.go
+++ b/check-disk/lib/check-disk.go
@@ -276,12 +276,12 @@ func filterPartitionsByInclusion(partitions []gpud.PartitionStat, list []string,
 	for _, partition := range partitions {
 		var ok = false
 		for _, l := range list {
-			if (l == key(partition)) {
+			if l == key(partition) {
 				ok = true
 				break
 			}
 		}
-		if (ok) {
+		if ok {
 			newPartitions = append(newPartitions, partition)
 		}
 	}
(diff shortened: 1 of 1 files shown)


check-disk/lib/check-disk.go
```diff
@@ -46,17 +46,13 @@ func checkStatus(current checkers.Status, threshold string, units float64, disk
...
-			inodesFreePct = float64(0)
-		}
 
-		if chkInode && v > inodesFreePct {
+		if chkInode && (disk.InodesTotal != 0 && v > inodesFreePct) {
 			current = status
 		}
 
-		if !chkInode && (v > freePct || v > inodesFreePct) {
+		if !chkInode && (v > freePct || (disk.InodesTotal != 0 && v > inodesFreePct)) {
```
susisu: I don't think comparing `inodesFreePct` here is correct behavior, since the help says `Exit with WARNING status if less than N units or N% of disk are free`, not N% of inodes.

check-disk/lib/check-disk.go
```diff
@@ -46,17 +46,13 @@ func checkStatus(current checkers.Status, threshold string, units float64, disk
...
 		freePct := float64(100) - disk.UsedPercent
-
 		inodesFreePct := float64(100) - disk.InodesUsedPercent
-		if float64(disk.InodesTotal) == float64(0) {
-			inodesFreePct = float64(0)
-		}
 
-		if chkInode && v > inodesFreePct {
+		if chkInode && (disk.InodesTotal != 0 && v > inodesFreePct) {
 			current = status
```
shibayu36: I did not understand what disk.InodesTotal! = 0 means for just a moment.  So How about writing comment, for example `If InodesTotal == 0, the system don't use inode`

astj:
This change itself looks good.

check-disk/lib/check-disk.go
susisu: I will propose to change this in another PR. Thanks!

check-disk/lib/check-disk.go
```diff
@@ -46,17 +46,13 @@ func checkStatus(current checkers.Status, threshold string, units float64, disk
...
 
 		freePct := float64(100) - disk.UsedPercent
-
 		inodesFreePct := float64(100) - disk.InodesUsedPercent
-		if float64(disk.InodesTotal) == float64(0) {
-			inodesFreePct = float64(0)
-		}
 
-		if chkInode && v > inodesFreePct {
+		if chkInode && (disk.InodesTotal != 0 && v > inodesFreePct) {
```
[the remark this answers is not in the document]
susisu: renamed in bea07d5 :+1:

check-disk/lib/check-disk.go
susisu: I added a note 94e20dc :+1:

PR #266, opened 2018-12-17
[check-disk] Do not check inodes% along with disk%

susisu:
As mentioned [here](https://github.com/mackerelio/go-check-plugins/pull/265#discussion_r241683338), checking inodes% when `-w N%` / `-c N%` specified is not preferable behavior, as the help does not refer to it.
This PR fixes it so that one can use `-w` / `-c` for disk% and `-W` / `-K` for inodes% separately.

### DIFF
--- a/check-disk/lib/check-disk.go
+++ b/check-disk/lib/check-disk.go
@@ -53,7 +53,7 @@ func checkStatus(current checkers.Status, threshold string, units float64, disk 
 			current = status
 		}
 
-		if !chkInode && (thresholdPct > freePct || (disk.InodesTotal != 0 && thresholdPct > inodesFreePct)) {
+		if !chkInode && (thresholdPct > freePct) {
 			current = status
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


check-disk/lib/check-disk.go
```diff
@@ -75,6 +64,29 @@ func checkStatus(current checkers.Status, threshold string, units float64, disk
...
+	}
+
+	thresholdPct, err := strconv.ParseFloat(strings.TrimRight(threshold, "%"), 64)
+	if err != nil {
+		return checkers.UNKNOWN, err
+	}
+
+	// Skip checking if disk.InodesTotal == 0 since inodesFreePct is meaningless.
+	if disk.InodesTotal == 0 {
+		return current, nil
```
susisu: I kept the previous behavior (just skipping checks), but perhaps it would be better to return UNKNOWN status.

susisu: What do you think? @astj

susisu:
Thank you!

PR #268, opened 2018-12-28
[check-log]stop reading logs on SIGTERM

lufia:
I added signal handler to check-log plugin to catch timeout from mackerel-agent.

When *check-log* plugin catches SIGTERM (Unix) or `os.Interrupt` (Windows), the plugin will read bytes until next newline and stop reading logs, and respond result to Mackerel.

### DIFF
--- a/check-log/lib/check-log.go
+++ b/check-log/lib/check-log.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -22,6 +23,9 @@ import (
 	"github.com/mattn/go-zglob"
 	enc "golang.org/x/text/encoding"
 )
+
+// overwritten with syscall.SIGTERM on unix environment (see check-log_unix.go)
+var defaultSignal = os.Interrupt
 
 type logOpts struct {
 	LogFile             string   `short:"f" long:"file" value-name:"FILE" description:"Path to log file"`
@@ -46,6 +50,8 @@ type logOpts struct {
 	fileListFromPattern []string
 	origArgs            []string
 	decoder             *enc.Decoder
+
+	testHookNewBufferedReader func(r io.Reader) *bufio.Reader
 }
 
 func (opts *logOpts) prepare() error {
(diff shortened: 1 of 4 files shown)


check-log/lib/check-log.go
```diff
@@ -316,9 +332,18 @@ func (opts *logOpts) searchLog(logFile string) (int64, int64, string, error) {
...
+}
+
+func (opts *logOpts) searchReader(ctx context.Context, rdr io.Reader) (warnNum, critNum, readBytes int64, errLines string, err error) {
+	newReader := opts.testHookNewBufferedReader
+	if newReader == nil {
+		newReader = newBufferedReader
+	}
+
+	r := newReader(rdr)
+	for ctx.Err() == nil {
```
astj: With this implementation, searching a file is not interrupted after this `ctx.Err()` check.
It's better to stop scanning on context cancel (= signal), even after this check.

check-log/lib/check-log.go
```diff
@@ -220,7 +236,7 @@ func run(args []string) *checkers.Checker {
 	return checkers.NewChecker(checkSt, msg)
 }
 
-func (opts *logOpts) searchLog(logFile string) (int64, int64, string, error) {
+func (opts *logOpts) searchLog(ctx context.Context, logFile string) (int64, int64, string, error) {
 	stateFile := getStateFile(opts.StateDir, logFile, opts.origArgs)
```
astj: Maybe we should return without scanning when ctx has been already canceled at start of this method.

check-log/lib/check-log.go
```diff
@@ -171,7 +187,7 @@ func run(args []string) *checkers.Checker {
 			missingFiles = append(missingFiles, f)
```
astj: We should check whether `ctx` is alive in this `range` loop so that we can avoid extra `os.Stat` for files after `ctx` is over.

check-log/lib/check-log.go
astj: Oops, I misread `for` as `if`...

lufia:
@astj I changed PR to avoid extra syscalls.

lufia:
Added a log like:

```
2019/01/16 12:06:50 check-log is exiting: caught a signal: terminated
```

PR #270, opened 2019-01-15
 [check-log][check-windows-eventlog] Improve atomic write

itchyny:
This pull request introduces https://github.com/natefinch/atomic library. The library is simple enough and fixes the problem os.Rename is not atomic on Windows. Also reduces one remove call compared to our writeFileAtomically function.


Songmu:
LGTM

itchyny:
Thank you.

PR #271, opened 2019-01-29
add options for client auth and adding to ca bundle

ephur:
add functionality to allow using client cert authentication, and providing a custom CA cert to add to the system CA bundle.

### DIFF
--- a/check-http/lib/check_http.go
+++ b/check-http/lib/check_http.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -16,7 +17,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/jessevdk/go-flags"
+	flags "github.com/jessevdk/go-flags"
 	"github.com/mackerelio/checkers"
 )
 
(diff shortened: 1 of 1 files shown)


mattn:
Go 1.8 had implementation([I added](https://go-review.googlesource.com/c/go/+/30578/1)) of x 509.System CertPool(), but it was reverted since Windows doesn't ship with bundling all of trusted root certificates. Windows downloads them on demand. Go team is still working on this. https://github.com/golang/go/issues/16736

ephur:
> @ephur Thank you for your contribution.
> 
> Go 1.11 on Windows, `x509.SystemCertPool()` always returns an error. So, on Windows, `tls.Config.RootCAs` will set with value of `x509.NewCertPool()`.
> 
> I'm not sure whether this is same behavior as current implementation. Do you know?

Based on the follow up response and the upstream issue, I'm not sure the best way to move forward with this. Potentially the options could be disabled if windows is the detected platform, or perhaps not extending the systems cert pool, and instead just replace it only if that option is set. I didn't go this route initially because it meant more code and I was trying to keep it slim and to the point but to make it supported cross platform it seems like this may be required.

lufia:
@ephur 
Could you change to set `tls.Config.RootCAs` to nil unless `-cacert` is set?
And could you add a note into README that `-cacert` is not working on Windows?

On Windows, if `tls.Config.RootCAs` set value of `x509.NewCertPool()`, verifying certificates will fail with **x509: certificate signed by unknown authority**.

https://play.golang.org/p/gYKr63skv5S

I think it is no problem that new options can be available on all platforms if they limitations are documented well.

lufia:
@ephur would you like me to change this p-r?

itchyny:
Closing due to inactivity.

PR #273, opened 2019-02-10
added support for netbsd on check-load

paulbsd:


check-load/lib/check_load_netbsd.go
```diff
@@ -0,0 +1,31 @@
+package checkload
+
+import (
+	"os/exec"
+	"strconv"
+	"strings"
+)
+
+func getloadavg() (loadavgs [3]float64, err error) {
+	outputBytes, err := exec.Command("sysctl", "-n", "vm.loadavg").Output()
```
astj: Using `unix.SysctlRaw` is better if possible, to avoid overhead of executing command.
(ref: https://github.com/mackerelio/go-osstat/blob/master/loadavg/loadavg_bsd_nocgo.go )

astj:
Thank you for contributing!

Looks good, but, as commented above, using `unix.SysctlRaw` is better if possible.  Or is there any reason to use `exec.Command` ?

paulbsd:
Thanks for feedback.
I commited another change, using github.com/mackerelio/go-osstat/loadavg, which use unix.SysctlRaw.
May go-osstat be used on any other OS ?

astj:
> May go-osstat be used on any other OS for check-load ?

Yes. Though we should remember that go-osstat uses cgo, but in case of loadavg it has fallback to non-cgo so there's no problem!

astj:
@paulbsd 
I'd like to merge this p-r for now. If you're interested in using go-osstat more widely, I'll appreciate it in another pull request. Thank you again for contribution!

PR #276, opened 2019-02-18
Add NTP stratum check

EijiSugiura:
Add ability to send warning, when NTP client failed to reach its stratum.

### DIFF
--- a/check-ntpoffset/lib/check-ntpoffset.go
+++ b/check-ntpoffset/lib/check-ntpoffset.go
@@ -236,15 +236,15 @@ func getNTPStratumFromNTPD() (stratum string, err error) {
 
 func parseNTPStratumFromNTPD(out io.Reader) (string, error) {
 	scr := bufio.NewScanner(out)
-        const stratumPrefix = "*"
+	const stratumPrefix = "*"
 	for scr.Scan() {
 		line := scr.Text()
 		if strings.HasPrefix(line, stratumPrefix) {
-                        flds := strings.Fields(line)
+			flds := strings.Fields(line)
 			if len(flds) != 10 {
 				return "unknown", fmt.Errorf("failed to get ntp stratum")
 			}
-                        return strings.TrimPrefix(flds[0], stratumPrefix), nil
+			return strings.TrimPrefix(flds[0], stratumPrefix), nil
 		}
 	}
 	return "unknown", fmt.Errorf("failed to get ntp stratum")
@@ -264,11 +264,11 @@ func parseNTPStratumFromChrony(out io.Reader) (string, error) {
 	for scr.Scan() {
 		line := scr.Text()
 		if strings.HasPrefix(line, stratumPrefix) {
-                        flds := strings.Fields(line)
+			flds := strings.Fields(line)
 			if len(flds) != 10 {
 				return "unknown", fmt.Errorf("failed to get ntp stratum")
 			}
-                        return flds[1], nil
+			return flds[1], nil
 		}
 	}
 	return "unknown", fmt.Errorf("failed to get ntp stratum")
--- a/check-ntpoffset/lib/check-ntpoffset.go
+++ b/check-ntpoffset/lib/check-ntpoffset.go
@@ -51,11 +51,6 @@ func run(args []string) *checkers.Checker {
 
 	var chkSt checkers.Status
 	var msg string
-	stratum, err := getNTPStratum()
-	if err != nil {
-		msg = fmt.Sprintf("ntp stratum not found offset is %f(actual), %f(warning threshold), %f(critical threshold)", math.Abs(offset), opts.Warn, opts.Crit)
-		return checkers.NewChecker(checkers.WARNING, msg)
-	}
 
 	if opts.Crit < math.Abs(offset) {
 		msg = fmt.Sprintf("ntp offset is over %f(actual) > %f(threshold)", math.Abs(offset), opts.Crit)
@@ -64,7 +59,7 @@ func run(args []string) *checkers.Checker {
 		msg = fmt.Sprintf("ntp offset is over %f(actual) > %f(threshold)", math.Abs(offset), opts.Warn)
 		chkSt = checkers.WARNING
 	} else {
-		msg = fmt.Sprintf("ntp stratum=%s, offset is %f(actual) < %f(warning threshold), %f(critial threshold)", stratum, math.Abs(offset), opts.Warn, opts.Crit)
+		msg = fmt.Sprintf("ntp offset is %f(actual) < %f(warning threshold), %f(critial threshold)", math.Abs(offset), opts.Warn, opts.Crit)
 		chkSt = checkers.OK
 	}
 
@@ -166,7 +161,7 @@ func getNTPOffsetFromNTPServers(ntpServers string) (offset float64, err error) {
 }
 
 func getNTPOffsetFromNTPD() (offset float64, err error) {
-	err = withCmd(exec.Command(cmdNTPq, "-c", "rv 0 offset"), func(out io.Reader) error {
+	err = withCmd(exec.Command(cmdNTPq, "-c", "rv 0 stratum,offset"), func(out io.Reader) error {
 		offset, err = parseNTPOffsetFromNTPD(out)
 		return err
 	})
@@ -175,11 +170,32 @@ func getNTPOffsetFromNTPD() (offset float64, err error) {
 
 func parseNTPOffsetFromNTPD(out io.Reader) (float64, error) {
 	scr := bufio.NewScanner(out)
+	const stratumPrefix = "stratum="
+	var stratumErr error
 	const offsetPrefix = "offset="
 	for scr.Scan() {
 		line := scr.Text()
-		if strings.HasPrefix(line, offsetPrefix) {
-			return strconv.ParseFloat(strings.TrimPrefix(line, offsetPrefix), 64)
+		for _, column := range strings.Split(line, ",") {
+			column = strings.TrimPrefix(column, " ")
+			if strings.HasPrefix(column, stratumPrefix) {
+				var stratum int64
+				stratum, stratumErr = strconv.ParseInt(strings.TrimPrefix(column, stratumPrefix), 10, 64)
+				if stratumErr != nil {
+					return 0.0, stratumErr
+				}
+				// the stratum of the remote machine. 16 is "unsynchronized"
+				// ref. https://support.ntp.org/bin/view/Support/TroubleshootingNTP
+				if stratum == 16 {
+					stratumErr = fmt.Errorf("not synchronized to stratum")
+				}
+			}
+			if strings.HasPrefix(column, offsetPrefix) {
+				offset, err := strconv.ParseFloat(strings.TrimPrefix(column, offsetPrefix), 64)
+				if err != nil {
+					return 0.0, err
+				}
+				return offset, stratumErr
+			}
 		}
 	}
 	return 0.0, fmt.Errorf("couldn't get ntp offset. ntpd process may be down")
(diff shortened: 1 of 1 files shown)


susisu:
Thank you for the contribution!

Is this still work in progress?
If not so, can you format the code using `gofmt` and fix [linter warnings](https://travis-ci.org/mackerelio/go-check-plugins/builds/494957098#L678-L679), please?

EijiSugiura:
Thank you for informing me!

I've fixed around lint warnings.
Could you please review this?

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -51,14 +51,20 @@ func run(args []string) *checkers.Checker {
 
 	var chkSt checkers.Status
 	var msg string
+	stratum, err := getNTPStratum()
+	if err != nil {
+		msg = fmt.Sprintf("ntp stratum not found offset is %f(actual), %f(warning threshold), %f(critical threshold)", math.Abs(offset), opts.Warn, opts.Crit)
+		return checkers.NewChecker(checkers.WARNING, msg)
```
susisu: I think returning `UNKNOWN` result is better for this case, because the main purpose of the program (checking NTP offset) is not accomplished.

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -206,3 +211,65 @@ func parseNTPOffsetFromChrony(out io.Reader) (offset float64, err error) {
 	}
 	return 0.0, fmt.Errorf("failed to get ntp offset")
 }
+
+func getNTPStratum() (string, error) {
+	ntpdName, err := detectNTPDname()
```
susisu: `detectNTPDName` is also called in `getNTPOffset`.
Can this be called only once?

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -51,14 +51,20 @@ func run(args []string) *checkers.Checker {
 
 	var chkSt checkers.Status
 	var msg string
+	stratum, err := getNTPStratum()
+	if err != nil {
```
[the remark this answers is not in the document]
susisu: Isn't it better to check stratum *before* retrieving offset?

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -51,14 +51,20 @@ func run(args []string) *checkers.Checker {
...
+
 	if opts.Crit < math.Abs(offset) {
 		msg = fmt.Sprintf("ntp offset is over %f(actual) > %f(threshold)", math.Abs(offset), opts.Crit)
 		chkSt = checkers.CRITICAL
 	} else if opts.Warn < math.Abs(offset) {
 		msg = fmt.Sprintf("ntp offset is over %f(actual) > %f(threshold)", math.Abs(offset), opts.Warn)
 		chkSt = checkers.WARNING
 	} else {
-		msg = fmt.Sprintf("ntp offset is %f(actual) < %f(warning threshold), %f(critial threshold)", math.Abs(offset), opts.Warn, opts.Crit)
+		msg = fmt.Sprintf("ntp stratum=%s, offset is %f(actual) < %f(warning threshold), %f(critial threshold)", stratum, math.Abs(offset), opts.Warn, opts.Crit)
```
susisu: Should the stratum name be added to the other messages above?

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -206,3 +211,65 @@ func parseNTPOffsetFromChrony(out io.Reader) (offset float64, err error) {
 	}
 	return 0.0, fmt.Errorf("failed to get ntp offset")
 }
+
+func getNTPStratum() (string, error) {
+	ntpdName, err := detectNTPDname()
+	if err != nil {
+		return "unknown", err
```
susisu: I think it is sufficient to return an empty string `""` instead of `"unknown"` when error occurred, because it will not be used at all.

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -206,3 +211,65 @@ func parseNTPOffsetFromChrony(out io.Reader) (offset float64, err error) {
 	}
 	return 0.0, fmt.Errorf("failed to get ntp offset")
 }
+
+func getNTPStratum() (string, error) {
```
susisu: What should this function do when `ntpServers` are specified?
Is it OK to get stratum from ntpd in such case?

EijiSugiura: I suppose nothing to do for that case, because `ntpServers` are equal to `stratum`.

check-ntpoffset/lib/check-ntpoffset.go
EijiSugiura: I prefer to return `WARNING`. There are 3 cases that we should care.

1.  `offset` is able to retrieved, and synchronized to stratum
2. `offset` is able to retrieved, but failed to synchronized to stratum
3. failed to get `offset`

`WARNING` is suitable for case2, so that NTP stratum could be not synchronized but accessible state.

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -206,3 +211,65 @@ func parseNTPOffsetFromChrony(out io.Reader) (offset float64, err error) {
...
+	err = withCmd(exec.Command(cmdNTPq, "-np"), func(out io.Reader) error {
+		stratum, err = parseNTPStratumFromNTPD(out)
+		return err
+	})
+	return stratum, err
+}
+
+func parseNTPStratumFromNTPD(out io.Reader) (string, error) {
+	scr := bufio.NewScanner(out)
+        const stratumPrefix = "*"
```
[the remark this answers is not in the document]
EijiSugiura: Sorry for inconvenience, I fixed it.

check-ntpoffset/lib/check-ntpoffset.go
EijiSugiura: Thank you for pointing it out.
I've replace `unknown` with  empty string.

check-ntpoffset/lib/check-ntpoffset.go
EijiSugiura: But for simplify the problem, I agree you.

check-ntpoffset/lib/check-ntpoffset.go
EijiSugiura: I've changed  to get stratum and offset with one command only.

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -161,7 +161,7 @@ func getNTPOffsetFromNTPServers(ntpServers string) (offset float64, err error) {
 }
 
 func getNTPOffsetFromNTPD() (offset float64, err error) {
-	err = withCmd(exec.Command(cmdNTPq, "-c", "rv 0 offset"), func(out io.Reader) error {
+	err = withCmd(exec.Command(cmdNTPq, "-c", "rv 0 stratum,offset"), func(out io.Reader) error {
```
EijiSugiura: To check offset and stratum with one commend.

check-ntpoffset/lib/check-ntpoffset.go
susisu: I am concerning about that, if it returns WARNING for the case 2, interpretation of WARNING result would be ambiguous.
What do you think?

check-ntpoffset/lib/check-ntpoffset.go
EijiSugiura: I removed the log message about stratum.

check-ntpoffset/lib/check-ntpoffset.go
EijiSugiura: For now, returns `UNKNOWN` in case failed to get NTP stratum.

check-ntpoffset/lib/check-ntpoffset.go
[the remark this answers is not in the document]
EijiSugiura: now, we can get offset and stratum with a single command.

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -170,11 +170,32 @@ func getNTPOffsetFromNTPD() (offset float64, err error) {
...
+				var stratum int64
+				stratum, stratumErr = strconv.ParseInt(strings.TrimPrefix(column, stratumPrefix), 10, 64)
+				if stratumErr != nil {
+					return 0.0, stratumErr
+				}
+				// the stratum of the remote machine. 16 is "unsynchronized"
+				// ref. https://support.ntp.org/bin/view/Support/TroubleshootingNTP
+				if stratum == 16 {
+					stratumErr = fmt.Errorf("not synchronized to stratum")
+				}
```
susisu: How about returning `stratumErr` here?

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -189,9 +210,23 @@ func getNTPOffsetFromChrony() (offset float64, err error) {
...
+			if len(flds) != 3 {
+				return 0.0, fmt.Errorf("failed to get ntp stratum")
+			}
+			var stratum int64
+			stratum, stratumErr = strconv.ParseInt(flds[2], 10, 64)
+			// Documents does NOT indicate about "Stratum : 0" means unsychronized...
+			// https://chrony.tuxfamily.org/manual.html#tracking-command
+			if stratum == 0 && stratumErr == nil {
+				stratumErr = fmt.Errorf("not synchronized to stratum")
+			}
```
susisu: Ditto.

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -170,11 +170,32 @@ func getNTPOffsetFromNTPD() (offset float64, err error) {
...
-			return strconv.ParseFloat(strings.TrimPrefix(line, offsetPrefix), 64)
+		for _, column := range strings.Split(line, ",") {
+			column = strings.TrimPrefix(column, " ")
+			if strings.HasPrefix(column, stratumPrefix) {
+				var stratum int64
+				stratum, stratumErr = strconv.ParseInt(strings.TrimPrefix(column, stratumPrefix), 10, 64)
+				if stratumErr != nil {
+					return 0.0, stratumErr
+				}
+				// the stratum of the remote machine. 16 is "unsynchronized"
```
susisu: `stratum` here is the stratum of **this computer**, isn't it?

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -189,9 +210,23 @@ func getNTPOffsetFromChrony() (offset float64, err error) {
...
 	for scr.Scan() {
 		line := scr.Text()
+		if strings.HasPrefix(line, "Stratum") {
+			flds := strings.Fields(line)
+			if len(flds) != 3 {
+				return 0.0, fmt.Errorf("failed to get ntp stratum")
+			}
+			var stratum int64
+			stratum, stratumErr = strconv.ParseInt(flds[2], 10, 64)
+			// Documents does NOT indicate about "Stratum : 0" means unsychronized...
```
susisu: So `Stratum : 0` actually means that the computer is unsynchronized.
(I know, of course, stratum-0 should be the most reliable one, but it is OK because it is indeed "unsynchronized".)

susisu:
[As I said before](https://github.com/mackerelio/go-check-plugins/pull/276#discussion_r259242339), this PR introduces a breaking change, especially for users who do not care about whether their computers are synchronized or not (I'm not sure that is good, though).
How about adding a flag to opt in the stratum check?

susisu:
I opened a new PR #278 where I added some changes (refactoring and tests), so I would like to close this PR.

PR #277, opened 2019-02-20
[check-http] add --proxy option

astj:
Although check-http can request using proxy via `HTTP_PROXY` environment variable, imo it's useful when we can specify the proxy url from command line option.

### DIFF
--- a/check-http/README.md
+++ b/check-http/README.md
@@ -73,6 +73,7 @@ To request via proxy (http/https/socks5)
 ```shell
 check-http --proxy=http://localhost:8080 -u http://example.com # request via http://localhost:8080
 HTTP_PROXY=http://localhost:8080 check-http  -u http://example.com # Same. you can set proxy via environment variable
+check-http --proxy=http://user:pass@localhost:8080 -u http://example.com # basic authentication is also supported
 ```
 ## For more information
 
--- a/check-http/README.md
+++ b/check-http/README.md
@@ -50,6 +50,7 @@ command = ["check-http", "-u", "http://example.com"]
   -p, --pattern=                              Expected pattern in the content
       --max-redirects=                        Maximum number of redirects followed (default: 10)
       --connect-to=HOST1:PORT1:HOST2:PORT2    Request to HOST2:PORT2 instead of HOST1:PORT1
+  -x, --proxy=[PROTOCOL://]HOST[:PORT]        Use the specified proxy. PROTOCOL's default is http, and PORT's default is 1080.
 ```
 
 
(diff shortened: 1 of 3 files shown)


check-http/lib/check_http.go
```diff
@@ -30,6 +31,7 @@ type checkHTTPOpts struct {
 	Regexp             string   `short:"p" long:"pattern" description:"Expected pattern in the content"`
 	MaxRedirects       int      `long:"max-redirects" description:"Maximum number of redirects followed" default:"10"`
 	ConnectTos         []string `long:"connect-to" value-name:"HOST1:PORT1:HOST2:PORT2" description:"Request to HOST2:PORT2 instead of HOST1:PORT1"`
+	Proxy              string   `short:"x" long:"proxy" value-name:"[PROTOCOL://]HOST[:PORT]" description:"Use the specified proxy. PROTOCOL's default is http, and PORT's default is 1080."`
```
mattn: [nit] Should add note that can pass `http://user:password@proxy.com` ? (but it's long-winded)

check-http/README.md
```diff
@@ -50,6 +50,7 @@ command = ["check-http", "-u", "http://example.com"]
   -p, --pattern=                              Expected pattern in the content
       --max-redirects=                        Maximum number of redirects followed (default: 10)
       --connect-to=HOST1:PORT1:HOST2:PORT2    Request to HOST2:PORT2 instead of HOST1:PORT1
+  -x, --proxy=[PROTOCOL://]HOST[:PORT]        Use the specified proxy. PROTOCOL's default is http, and PORT's default is 1080.
```
syou6162: Is there any reasons why you select `1080` as default port?

astj: Those default values are taken from `curl`.
```
       -x, --proxy [protocol://]host[:port]
              Use the specified proxy.

              The  proxy  string  can  be  specified with a protocol:// prefix. No protocol specified or http:// will be treated as HTTP proxy. Use socks4://, socks4a://, socks5:// or
              socks5h:// to request a specific SOCKS version to be used.  (The protocol support was added in curl 7.21.7)

              HTTPS proxy support via https:// protocol prefix was added in 7.52.0 for OpenSSL, GnuTLS and NSS.

              Unrecognized and unsupported proxy protocols cause an error since 7.52.0.  Prior versions may ignore the protocol and use http:// instead.

              If the port number is not specified in the proxy string, it is assumed to be 1080.

              This option overrides existing environment variables that set the proxy to use. If there's an environment variable setting a proxy, you can set proxy to ""  to  override
              it.

              All  operations that are performed over an HTTP proxy will transparently be converted to HTTP. It means that certain protocol specific operations might not be available.
              This is not the case if you can tunnel through the proxy, as one with the -p, --proxytunnel option.

              User and password that might be provided in the proxy string are URL decoded by curl. This allows you to pass in special characters such as @ by using %40 or pass  in  a
              colon with %3a.

              The proxy host can be specified the exact same way as the proxy environment variables, including the protocol prefix (http://) and the embedded user + password.

              If this option is used several times, the last one will be used.
```

check-http/lib/check_http.go
astj: Hmm, command line description is so narrow to describe everything, but adding `[USER:PASS@` seems acceptable.
Thank you for suggestion. I added at description (and tests for safe) at c92b8564a2a1d1df352e8d8bc4bf63a54ebb09af

PR #278, opened 2019-03-12
Add NTP stratum check to check-ntpoffset

susisu:
From #276.

In addition to the original PR, I made some changes:

- refactor and fix parse functions
- add a command-line flag `--check-stratum`
- add test cases

### DIFF
--- a/check-ntpoffset/lib/check-ntpoffset.go
+++ b/check-ntpoffset/lib/check-ntpoffset.go
@@ -199,7 +199,7 @@ func parseNTPOffsetFromNTPD(out io.Reader, checkStratum bool) (float64, error) {
 		}
 	}
 	if offset == nil {
-		return 0.0, fmt.Errorf("couldn't get ntp offset. ntpd process may be down")
+		return 0.0, fmt.Errorf("failed to get ntp offset")
 	}
 	return *offset, nil
 }
--- a/check-ntpoffset/lib/check-ntpoffset.go
+++ b/check-ntpoffset/lib/check-ntpoffset.go
@@ -18,10 +18,11 @@ import (
 )
 
 var opts struct {
-	Crit       float64 `short:"c" long:"critical" default:"100" description:"Critical threshold of ntp offset(ms)"`
-	Warn       float64 `short:"w" long:"warning" default:"50" description:"Warning threshold of ntp offset(ms)"`
-	NTPServers string  `short:"s" long:"ntp-servers" default:"" description:"Use specified NTP Servers(plural servers can be set separated by ,). When set plural servers, use first response. If not set, use local command just like ntpd/chronyd."`
-	NTPTimeout int     `short:"t" long:"ntp-timeout" default:"15" description:"Timeout of NTP Server Querying(in seconds)."`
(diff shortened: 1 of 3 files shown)


check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -18,10 +18,11 @@ import (
...
-	NTPServers string  `short:"s" long:"ntp-servers" default:"" description:"Use specified NTP Servers(plural servers can be set separated by ,). When set plural servers, use first response. If not set, use local command just like ntpd/chronyd."`
-	NTPTimeout int     `short:"t" long:"ntp-timeout" default:"15" description:"Timeout of NTP Server Querying(in seconds)."`
+	Crit         float64 `short:"c" long:"critical" default:"100" description:"Critical threshold of ntp offset(ms)"`
+	Warn         float64 `short:"w" long:"warning" default:"50" description:"Warning threshold of ntp offset(ms)"`
+	NTPServers   string  `short:"s" long:"ntp-servers" default:"" description:"Use specified NTP Servers(plural servers can be set separated by ,). When set plural servers, use first response. If not set, use local command just like ntpd/chronyd."`
+	NTPTimeout   int     `short:"t" long:"ntp-timeout" default:"15" description:"Timeout of NTP Server Querying(in seconds)."`
+	CheckStratum bool    `short:"S" long:"check-stratum" description:"Check stratum and fail if the machine is not synchronized."`
```
susisu: I'm not really sure, but perhaps it is better to always enable this.

EijiSugiura:
Thank you for fixing!

Without `--check-stratum`, `check-ntpoffset` reports **OK** as below, even if the host does not synchronized to stratum.
```
NTP OK: ntp offset is 0.000000(actual) < 50.000000(warning threshold), 100.000000(critial threshold)
```

We were often faced with this situation, because we've restricted the outbound access by NTP and DNS. 
With `--check-stratum`, we can properly handle it.
As a side effect, "not synchronized error" will be happened just after ntpd/chronyd restarted.

check-ntpoffset/lib/check-ntpoffset.go
```diff
@@ -160,49 +161,94 @@ func getNTPOffsetFromNTPServers(ntpServers string) (offset float64, err error) {
 	}
 }
 
-func getNTPOffsetFromNTPD() (offset float64, err error) {
-	err = withCmd(exec.Command(cmdNTPq, "-c", "rv 0 offset"), func(out io.Reader) error {
-		offset, err = parseNTPOffsetFromNTPD(out)
+func getNTPOffsetFromNTPD(checkStratum bool) (offset float64, err error) {
+	err = withCmd(exec.Command(cmdNTPq, "-c", "rv 0 stratum,offset"), func(out io.Reader) error {
```
itchyny: Here `stratum,` can be omitted if checkStratum is false, right?

itchyny:
LGTM!

check-ntpoffset/lib/check-ntpoffset.go
susisu: Yes, but I think it does not cost much and always specifying `stratum` is simpler.

susisu:
Thank you!

PR #279, opened 2019-03-16
[ntservice] Enable to specify service name to exclude from match

a-know:


check-ntservice/README.md
```diff
@@ -42,8 +42,9 @@ command = ["check-ntservice", "--service-name", "SERVICE_NAME"]
 ### Options
 
 ```
-  -s, --service-name= service name
-  -l, --list-service  list service
+  -s, --service-name=    service name
+  -E, --exclude-service= service name to exclude from matching
```
lufia: I think traditional commands such as *diff*, *tar*, or *du* use often `x|X` character for meaning *exclude* instead of `e|E`, or no short notation.

check-ntservice/lib/check_ntservice_test.go
```diff
@@ -54,3 +77,41 @@ func TestNtService(t *testing.T) {
...
+		},
+		{
+			casename:      "check about running service with exclude option",
+			cmdline:       []string{"-s", "service", "-E", "stopped"},
+			expectStatus:  checkers.OK,
+			expectMessage: "",
+		},
+	}
+
+	mockServiceState()
```
lufia: This test had better reset `getServiceStateFunc` to initial state. Because tests that run after this test could be able to break.

For example:

```go
originalFunc := getServiceStateFunc
defer func(){
    getServiceStateFunc = originalFunc
}()
mockServiceState()
```

a-know: That's true! I fix it, effeb49 .

check-ntservice/README.md
a-know: I was not certain the character, either. I follow your point, ee53b4b .

PR #284, opened 2019-04-01
[check-ldap] Update go-ldap to v3

nonylene:
Current 386 binary throws runtime error on my environment (Debian 8, x86_64).

```
$ /usr/bin/check-ldap -b 'dc=example,dc=com' -w 5 -c 10
2019/04/01 19:13:03 ldap: recovered panic in processMessages: runtime error: invalid memory address or nil pointer dereference
LDAP UNKNOWN: LDAP Result Code 200 "Network Error": ldap: response channel closed
```

https://github.com/go-ldap/ldap/pull/199 should fix this issue and this patch exists only in v3, so I updated go-ldap package.

### DIFF
--- a/check-ldap/lib/check_ldap.go
+++ b/check-ldap/lib/check_ldap.go
@@ -8,7 +8,7 @@ import (
 
 	"github.com/jessevdk/go-flags"
 	"github.com/mackerelio/checkers"
-	"gopkg.in/ldap.v2"
+	"gopkg.in/ldap.v3"
 )
 
 type checkLDAPOpts struct {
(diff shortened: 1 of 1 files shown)


check-ldap/lib/check_ldap.go
```diff
@@ -8,7 +8,7 @@ import (
 
 	"github.com/jessevdk/go-flags"
 	"github.com/mackerelio/checkers"
-	"gopkg.in/ldap.v2"
+	"gopkg.in/ldap.v3"
```
astj: :eyes: https://github.com/go-ldap/ldap/blob/migrating-v2-v3/MIGRATING_v2-v3.md

astj:
Thank you!

PR #290, opened 2019-06-07
check-gateway: uses arping to check a gateway

boyska:
this PR adds a new check command.

It works calling both `arp` and `arping` commands. In the near future I plan to improve it, removing the need for the `arp` command (which is very system-dependent).

There are many implementations of `arping`: the one that this plugin requires is `arping-th` (package `arping` on ubuntu)

### DIFF
--- a/check-gateway/README.md
+++ b/check-gateway/README.md
@@ -1,0 +1,35 @@
+# check-gateway
+
+## Description
+
+Check ICMP Ping connections with the specified host forcing a specified gateway
+
+Depends on arping-th
+
+## Synopsis
+```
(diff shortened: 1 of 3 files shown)


lufia:
@boyska Hi. Thank you for your pull request.

This repository is standard plugin collection maintained by mainly Mackerel team so we don't merge new plugins actively.
Instead we suggest to you add check-gateway plugin to plugin-registry repository.

- Plugin registry: https://github.com/mackerelio/plugin-registry

Plugins added above registry can install it easily with `mkr plugin install`, and can update it easily on your own repository.

- mkr: https://github.com/mackerelio/mkr

PR #294, opened 2019-07-11
[check-log] When specified multiple exclude pattern, perform search that excludes all conditions.

tukaelu:
This p-r is for excludes the lines that contain exclude strings in the `AND` condition when multiple `--exclude` options are specified.

### DIFF
--- a/check-log/lib/check-log.go
+++ b/check-log/lib/check-log.go
@@ -33,7 +33,7 @@ type logOpts struct {
 	LogFile             string   `short:"f" long:"file" value-name:"FILE" description:"Path to log file"`
 	Pattern             []string `short:"p" long:"pattern" required:"true" value-name:"PAT" description:"Pattern to search for. If specified multiple, they will be treated together with the AND operator"`
 	SuppressPattern     bool     `long:"suppress-pattern" description:"Suppress pattern display"`
-	Exclude             string   `short:"E" long:"exclude" value-name:"PAT" description:"Pattern to exclude from matching"`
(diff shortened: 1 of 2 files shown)


check-log/lib/check-log.go
```diff
@@ -403,10 +403,21 @@ func (opts *logOpts) searchReader(ctx context.Context, rdr io.Reader) (warnNum,
...
 	var matches []string
 	for _, pReg := range opts.patternReg {
-		eReg := opts.excludeReg
 
 		matches = pReg.FindStringSubmatch(line)
-		if len(matches) == 0 || (eReg != nil && eReg.MatchString(line)) {
+		if len(matches) != 0 {
+			if len(opts.excludeReg) > 0 {
+				excludeAllMatched := true
+				for _, eReg := range opts.excludeReg {
```
susisu: This nested loop may be avoided like this:
``` go
func (opts *logOpts) match(line string) (bool, []string) {
	var matches []string
	for _, pReg := range opts.patternReg {
		matches = pReg.FindStringSubmatch(line)
		if len(matches) == 0 {
			return false, nil
		}
	}
	if len(opts.excludeReg) > 0 {
		exclude := true
		for _, eReg := range opts.excludeReg {
			if !eReg.MatchString(line) {
				exclude = false
				break
			}
		}
		if exclude {
			return false, nil
		}
	}
	return true, matches
}
```

tukaelu:
Sorry for the late response.
Since all corresponded, please confirm.

PR #296, opened 2019-08-04
Add check-ftp

ch1aki:
This plugin can be used check behave for ftp server.


ch1aki:
I read https://mackerel.io/ja/blog/entry/feature/20171116
I will try to register in the plugin registry!

PR #302, opened 2019-10-22
[check-log] Ignore broken/unexpected json on reading state

astj:
I'm not sure this is the best approach, but worth consider…

### DIFF
--- a/check-log/lib/check-log.go
+++ b/check-log/lib/check-log.go
@@ -283,6 +283,7 @@ func (opts *logOpts) searchLog(ctx context.Context, logFile string) (int64, int6
 	}
 
 	if !opts.NoState && !opts.CheckFirst {
+		// skip whole file if state file does not exist
 		if _, err = os.Stat(stateFile); os.IsNotExist(err) {
 			skipBytes = stat.Size()
 		}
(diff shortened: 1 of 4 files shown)


check-log/lib/check-log.go
```diff
@@ -444,6 +451,8 @@ func getStateFile(stateDir, f string, args []string) string {
 	)
 }
 
+var errValidStateFileNotFound = fmt.Errorf("state file not found, or corrupted")
+
 func getBytesToSkip(f string) (int64, error) {
 	state, err := loadState(f)
```
lufia: If state file is corrupted, `loadState()` will return `nil, nil`.
Is it intended behavior that `getBytesToSkipOld` will be called to try to read old state file in this case?

check-log/lib/check-log.go
```diff
@@ -247,11 +247,14 @@ func (opts *logOpts) searchLog(ctx context.Context, logFile string) (int64, int6
 		return 0, 0, "", nil
 	}
 	stateFile := getStateFile(opts.StateDir, logFile, opts.origArgs)
-	skipBytes, inode := int64(0), uint(0)
+	skipBytes, inode, isFirstCheck := int64(0), uint(0), false
 	if !opts.NoState {
 		s, err := getBytesToSkip(stateFile)
 		if err != nil {
-			return 0, 0, "", err
+			if err != errValidStateFileNotFound {
```
lufia: `getBytesToSkip` can return `errValidStateFileNotFound` only if `getBytesToSkipOld` is called.

check-log/lib/check-log.go
astj: I thought the behavior is not the best but acceptable.
But finally I decided not to read old state file the case ****

PR #306, opened 2019-12-19
refactor Makefile and update dependencies

lufia:
I've updated **Makefile** and others for support Go Modules, and upgraded libraries.


ne-sachirou:
🙄

PR #330, opened 2020-02-20
disable unquoting of go-flags

### LINKED ISSUE #327: InvalidParameterException occurs when a symbol is entered in filter-pattern in check-aws-cloudwatch-logs
If you include ":" in `filter-pattern (-p)` option of check-aws-cloudwatch-logs, `InvalidParameterException` will occur.

```bash
$ check-aws-cloudwatch-logs --log-group-name log_group_name -p "\"err:\""
CloudWatch Logs UNKNOWN: InvalidParameterException:
```

Works fine with aws cli
```bash
$ aws logs filter-log-events --log-group-name log_group_name --filter-pattern "\"err:\""
```

This seems to be a matter of interpretation when `github.com/jessevdk/go-flags` contains`\"\"` in the string parameter.

lufia:
Fixes #327

go-flags package unquote options by default. This behavior was introduced at:
https://github.com/jessevdk/go-flags/pull/96

In the case of #327, `-p "\"err:\""` will be unquoted, then the same as `-p err:`. So I think `err:` expression might be going to violate the [Filter syntax](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html).

**Notes**
I fixed only for *check-aws-cloudwatch-logs* plugin. Other plugins is not fixed yet.

### DIFF
--- a/check-aws-cloudwatch-logs/lib/check-aws-cloudwatch-logs.go
+++ b/check-aws-cloudwatch-logs/lib/check-aws-cloudwatch-logs.go
@@ -23,13 +23,13 @@ import (
 )
 
 type logOpts struct {
-	LogGroupName        string `long:"log-group-name" required:"true" value-name:"LOG-GROUP-NAME" description:"Log group name"`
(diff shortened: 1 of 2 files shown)


masahide:
Thanks arigato!

PR #338, opened 2020-03-09
[check-redis] Add password support for check-redis

n-rodriguez:
Thank you!

### DIFF
--- a/check-redis/lib/check-redis.go
+++ b/check-redis/lib/check-redis.go
@@ -14,10 +14,11 @@ import (
 )
 
 type redisSetting struct {
-	Host    string `short:"H" long:"host" default:"localhost" description:"Hostname"`
-	Socket  string `short:"s" long:"socket" default:"" description:"Server socket"`
-	Port    string `short:"p" long:"port" default:"6379" description:"Port"`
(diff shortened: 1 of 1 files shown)


check-redis/lib/check-redis.go
```diff
@@ -63,6 +64,14 @@ func connectRedis(m redisSetting) (*redis.Client, error) {
 	if err != nil {
 		return nil, fmt.Errorf("couldn't connect: %s", err)
 	}
+
+	if m.Password != "" {
+		r := c.Cmd("AUTH", m.Password)
+		if r.Err != nil {
+			return nil, errors.New("couldn't execute query")
```
lufia: Do this *r.Err* contain the case of authentication failed by password incorrect? If so this error message is not kindly. I think it is better that an error includes an error from Redis, like `fmt.Errorf("couldn't authenticate: %v", r.Err)`.

n-rodriguez: >  I think it is better that an error includes an error from Redis

Right! Fixed :)

lufia:
Thank you! We will review this p-r.

n-rodriguez:
Thank you!

PR #365, opened 2020-05-15
[check-postgresql] add test.sh

lufia:
I added basic tests for check-postgresql.

### DIFF
--- a/check-postgresql/test.sh
+++ b/check-postgresql/test.sh
@@ -24,7 +24,7 @@ docker run -d \
 trap 'docker stop test-$plugin; docker rm test-$plugin; exit' EXIT
 sleep 10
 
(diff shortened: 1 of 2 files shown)


check-postgresql/test.sh
```diff
@@ -0,0 +1,32 @@
...
+user=postgres
+password=passpass
+docker run -d \
+	--name test-$plugin \
+	-p 15432:5432 \
+	-e POSTGRES_PASSWORD=$password postgres:11
+trap 'docker stop test-$plugin; docker rm test-$plugin; exit' EXIT
+sleep 10
+
+if $plugin --port 15432 --user=$user --password=$password >/dev/null 2>&1
```
yseto, line 27:
```suggestion
if $plugin connection --port 15432 --user=$user --password=$password >/dev/null 2>&1
```

PR #377, opened 2020-06-09
[check-ssl-cert] check intermediate- and root-certificates

lufia:
The intermediate- and root-certificates received from the server will be checked.

To check the raw certificates sent from the server:

```console
% openssl s_client -servername <domain> -connect  <domain>:443 -showcerts </dev/null >c.pem
% openssl crl2pkcs7 -nocrl -certfile c.pem | openssl pkcs7 -print_certs -text -noout
```


lufia:
On GitHub Checks, Travis's status is pending. But now, the check was completed successfully.

PR #403, opened 2020-08-19
[check-http]adding more options to check-http

fgouteroux:
Hi,

I want to add some other options for the check-http based on https://github.com/sensu-plugins/sensu-plugins-http/blob/master/bin/check-http.rb. Does it seems good ?

### DIFF
--- a/check-http/lib/check_http.go
+++ b/check-http/lib/check_http.go
@@ -38,7 +38,7 @@ type checkHTTPOpts struct {
 	RequireBytes       int64    `short:"B" long:"require-bytes" description:"Check the response contains exactly BYTES bytes" default:"-1"`
 	Body               string   `short:"d" long:"body" description:"Send a data body string with the request"`
 	MinBytes           int64    `short:"g" long:"min-bytes" description:"Check the response contains at least BYTES bytes" default:"-1"`
-	Timeout            int64    `short:"t" long:"timeout" description:"Set the total execution timeout in seconds" default:"15"`
+	Timeout            int64    `short:"t" long:"timeout" description:"Set the total execution timeout in seconds" default:"0"`
 	CertFile           string   `long:"cert-file" description:"A Cert file to use for client authentication"`
 	KeyFile            string   `long:"key-file" description:"A Key file to use for client authentication"`
 	CaFile             string   `long:"ca-file" description:"A CA Cert file to use for client authentication"`
@@ -297,8 +297,8 @@ func Run(args []string) *checkers.Checker {
 			return checkers.Unknown(err.Error())
 		}
 	} else {
-		buffer := bytes.NewBuffer([]byte(opts.Body))
-		req, err = http.NewRequest(opts.Method, opts.URL, buffer)
+		body := strings.NewReader(opts.Body)
+		req, err = http.NewRequest(opts.Method, opts.URL, body)
 		if err != nil {
 			return checkers.Unknown(err.Error())
 		}
--- a/check-http/lib/check_http.go
+++ b/check-http/lib/check_http.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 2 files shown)


lufia:
Hi @fgouteroux 
Sorry for late reply.

We welcome your pull requests!

But if there is standard way to do that, there are room to consider whether using the standard way or adding a option. For example, we might use *HTTP_PROXY* environment variable instead of `--proxy_url` option because Go refers *HTTP_PROXY* by default.

check-http/lib/check_http.go
```diff
@@ -245,22 +279,34 @@ func Run(args []string) *checkers.Checker {
...
-		return checkers.Unknown(err.Error())
+	var req *http.Request
+	if opts.Body == "" {
+		req, err = http.NewRequest(opts.Method, opts.URL, nil)
+		if err != nil {
+			return checkers.Unknown(err.Error())
+		}
+	} else {
+		buffer := bytes.NewBuffer([]byte(opts.Body))
+		req, err = http.NewRequest(opts.Method, opts.URL, buffer)
```
lufia, line 301: Is this better to use `strings.NewReader`?
https://golang.org/pkg/strings/#NewReader

check-http/lib/check_http.go
```diff
@@ -30,9 +31,17 @@ type checkHTTPOpts struct {
...
 	ConnectTos         []string `long:"connect-to" value-name:"HOST1:PORT1:HOST2:PORT2" description:"Request to HOST2:PORT2 instead of HOST1:PORT1"`
 	Proxy              string   `short:"x" long:"proxy" value-name:"[PROTOCOL://][USER:PASS@]HOST[:PORT]" description:"Use the specified proxy. PROTOCOL's default is http, and PORT's default is 1080."`
-        BasicAuth          string   `long:"user" value-name:"USER[:PASSWORD]" description:"Basic Authentication user ID and an optional password."`
+	BasicAuth          string   `long:"user" value-name:"USER[:PASSWORD]" description:"Basic Authentication user ID and an optional password."`
+	RequireBytes       int64    `short:"B" long:"require-bytes" description:"Check the response contains exactly BYTES bytes" default:"-1"`
+	Body               string   `short:"d" long:"body" description:"Send a data body string with the request"`
+	MinBytes           int64    `short:"g" long:"min-bytes" description:"Check the response contains at least BYTES bytes" default:"-1"`
+	Timeout            int64    `short:"t" long:"timeout" description:"Set the total execution timeout in seconds" default:"15"`
```
lufia, line 41: Is this default value comes from?

Basically, we shouldn't change current behavior so I suggest to change this value to zero, it means no timeout.
https://golang.org/pkg/net/http/#Client

fgouteroux: yes, you right, by defaut we should let 0.

fgouteroux:
> Hi @fgouteroux
> Sorry for late reply.
> 
> We welcome your pull requests!
> 
> But if there is standard way to do that, there are room to consider whether using the standard way or adding a option. For example, we might use _HTTP_PROXY_ environment variable instead of `--proxy_url` option because Go refers _HTTP_PROXY_ by default.

For which option, you think I can use environment variable ?

lufia:
Don't worry, that didn't mention to anything of this pull-request.

I attempted to tell general rules by above statement, it couldn't do well...

lufia:
@fgouteroux I reviewed this pull request, it will be merged.

I accept this changes. It will be merged later.

However I noticed that the commit 62cd556 and 1bc337c are same content, but they has different commit hash. Perhaps that commits appeared by `git merge`, or `git rebase`. Could I have you resolved that?
(I think it would resolve to set HEAD to 8d58134, the test of check-log fails occasionally on Windows. It is known issue.)

fgouteroux:
> @fgouteroux I reviewed this pull request, it will be merged.
> 
> I accept this changes. It will be merged later.
> 
> However I noticed that the commit [62cd556](https://github.com/mackerelio/go-check-plugins/commit/62cd5563137ff496d82077a3f2078fdcfce1bfec) and [1bc337c](https://github.com/mackerelio/go-check-plugins/commit/1bc337c1a9eca5291ad152d8b412d2d5c8ab87c1) are same content, but they has different commit hash. Perhaps that commits appeared by `git merge`, or `git rebase`. Could I have you resolved that?
> (I think it would resolve to set HEAD to [8d58134](https://github.com/mackerelio/go-check-plugins/commit/8d58134daabb4f1cd07f77e0a56feb36a255b117), the test of check-log fails occasionally on Windows. It is known issue.)

Yes, I think there is some changes for the test of check-log, so I rebase. Reverted now. Thanks for the review.

lufia (approved):
Thank you!

PR #405, opened 2020-08-24
fix check-load percpu output

fgouteroux:
Hi,

When using check-load `--percpu` option, the cpu load output is not divided per cpu. That could bring some misunderstood when define threshold warning/critical.

### DIFF
--- a/check-load/lib/check_load.go
+++ b/check-load/lib/check_load.go
@@ -63,11 +63,13 @@ func run(args []string) *checkers.Checker {
 		return checkers.Unknown(err.Error())
 	}
 
+	var thresholds_per_cpu []float64
 	result := checkers.OK
 	for i, load := range loadavgs {
 		if opts.PerCPU {
 			numCPU := runtime.NumCPU()
 			load = load / float64(numCPU)
+			thresholds_per_cpu = append(thresholds_per_cpu, load)
 		}
 		if load > cload[i] {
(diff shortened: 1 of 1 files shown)


check-load/lib/check_load.go
```diff
@@ -78,6 +80,11 @@ func run(args []string) *checkers.Checker {
 		}
 	}
 
-	msg := fmt.Sprintf("load average: %.2f, %.2f, %.2f", loadavgs[0], loadavgs[1], loadavgs[2])
+	var msg string
+	if opts.PerCPU {
+		msg = fmt.Sprintf("load average: %.2f, %.2f, %.2f", thresholdsPerCPU[0], thresholdsPerCPU[1], thresholdsPerCPU[2])
+	} else {
+		msg = fmt.Sprintf("load average: %.2f, %.2f, %.2f", loadavgs[0], loadavgs[1], loadavgs[2])
+	}
```
lufia, line 88: This block shows values that was used to check whether is higher than the threshould, so I think this could be removed the second `if opts.PerCPU` statement.

- add `load` into the slice, eg.named `threshould`, at the end of the `for ... range loads` loop.
- print `thresholds`.

fgouteroux: I follow your suggestion, thanks for the review.

lufia:
@fgouteroux Thank you!

PR #406, opened 2020-08-28
[check-log] stabilize time-dependent tests

astj:
`testCancel` sometimes fails on Windows (ref: https://github.com/mackerelio/go-check-plugins/pull/403#issuecomment-680782111).
I modified this test by removing timestamp comparison to avoid such failures.


lufia (approved):
awesome job!

PR #441, opened 2020-12-29
Closes #440 added sort the chec-disk output

### LINKED ISSUE #440: check-disk sort output by percentage
The order of the output really bothers me. I think the simplest is to sort by the used %.  
Ideally it would be sorted by what thresholds %, specific number or inode flavors. I think if this plugin as refactored it would be easy to implement this but at this time it just seemed to convoluted to me.

The ordering by used % should fit the majority of use cases.

hurrycaine:
Signed-off-by: hurrycaine <jason.e.joyner@gmail.com>

### DIFF
--- a/check-disk/lib/check-disk.go
+++ b/check-disk/lib/check-disk.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


ne-sachirou:
> The ordering by used % should fit the majority of use cases.

I agree with this. Thank you for this pull request.

ne-sachirou (approved):
🙆‍♀️

PR #456, opened 2021-02-01
Closes #455 check-load fix null pointer issue when critical

### LINKED ISSUE #455: check-load null pointer when critical
In this example the load average was around 1.3.

If the load was below critical then everything worked as it should.  Warning was fine, just Critical had the bug.

The issue was due to a break in a for loop inside the if critical block. This caused nothing to go into the threshold slice,
which used direct references to threshold[0],threshold[1],threshold[2] in the print statement.

I will provide a PR for the fix.

**Critical**
```
-bash-4.2$ ./check-load -w 1,1,1 -c 1.2,1.3,1.4 -r
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/mackerelio/go-check-plugins/check-load/lib.run(0xc00009a130, 0x5, 0x5, 0x0)
        E:/gowork/src/safetydance.premierinc.com/stash/tools/gosensuplugins/vendor/github.com/mackerelio/go-check-plugins/check-load/lib/check_load.go:83 +0x645
github.com/mackerelio/go-check-plugins/check-load/lib.Do()
        E:/gowork/src/safetydance.premierinc.com/stash/tools/gosensuplugins/vendor/github.com/mackerelio/go-check-plugins/check-load/lib/check_load.go:41 +0x68
main.main()
        E:/gowork/src/safetydance.premierinc.com/stash/tools/gosensuplugins/vendor/github.com/mackerelio/go-check-plugins/check-load/main.go:6 +0x20
```

**Warning**
```
-bash-4.2$ ./check-load -w 1,1,1 -c 2,2,3 -r
LOAD WARNING: load average: 1.33, 1.33, 1.33
```

**With FIx**
```
-bash-4.2$ ./check-load -w 1,1,1 -c 1.2,1.2,1.2 -r
LOAD CRITICAL: load average: 1.34, 1.34, 1.34


```

hurrycaine:
Moving the "if opts.PerCPU" into its own for loop makes the code cleaner in my opinion. There are 2 loops now but is is only a loop of 3.

It is important to go through all the loops to get the load divided by cpu (if that option selected). However, on the flip side it is important to not loop through all of the loops when you have reached a critical threshold.

This has to due to how the critical and warn are like this "-w 1,1,1 -c 2,2,2" and the actual value could be 2, 1.5, 1.  
In this scenario the 1st minute average is critical but if we keep looping 5 and 15 minute average would put it back to a Warning and we don't want that.  

We could have added an additional if critical (previous loop) before moving on to warning but it gets messy. 

I also pulled the runtime.NumCPU() outside of the loop as this should only be called once.

### DIFF
--- a/check-load/lib/check_load.go
+++ b/check-load/lib/check_load.go
@@ -63,23 +63,24 @@ func run(args []string) *checkers.Checker {
 		return checkers.Unknown(err.Error())
 	}
 
-	var thresholds []float64
-	result := checkers.OK
-	for i, load := range loadavgs {
-		if opts.PerCPU {
-			numCPU := runtime.NumCPU()
-			load = load / float64(numCPU)
+	if opts.PerCPU {
+		numCPU := runtime.NumCPU()
+		for i, load := range loadavgs {
+			loadavgs[i] = load / float64(numCPU)
 		}
-		if load > cload[i] {
+	}
+
+	result := checkers.OK
+	for i := range loadavgs {
+		if loadavgs[i] > cload[i] {
 			result = checkers.CRITICAL
 			break
 		}
-		if load > wload[i] {
+		if loadavgs[i] > wload[i] {
 			result = checkers.WARNING
 		}
-		thresholds = append(thresholds, load)
 	}
 
-	msg := fmt.Sprintf("load average: %.2f, %.2f, %.2f", thresholds[0], thresholds[1], thresholds[2])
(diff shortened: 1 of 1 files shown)


shibayu36 (approved):
Thanks for your pull request. LGTM!

shibayu36:
I've just merged and will be releasing it soon.

PR #464, opened 2021-02-25
[check-mysql] Use go-mysql-driver and sqlx instead of mymysql

nonylene:
This PR introduces go-mysql-driver and sqlx instead of mymysql to support a new password method `caching_sha2_password` and TLS certificats (TLS certificates is not yet supported in this PR).

In addition, I have made error logs more helpful and added some basic test scripts.


check-mysql/lib/check-mysql.go
```diff
@@ -52,12 +62,60 @@ SubCommands:`)
...
+}
+
+func getMySQLVersion(db *sql.DB) (*mysqlVersion, error) {
+	var rawVersion string
+	err := db.QueryRow("SELECT VERSION()").Scan(&rawVersion)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to query SELECT VERSION(): %s", err)
+	}
+
+	// Version example: mysql-8.0.1-dmr
```
astj, line 90: afaik, value from `SELECT VERSION()` (or underlying `show variables like 'version';`) is just like `8.0.23` or `5.5.44-0+deb8u1-log` (it's ancient one!), so this example does not represent those values?

check-mysql/lib/check-mysql.go
```diff
@@ -19,6 +22,13 @@ type mysqlSetting struct {
 	Pass   string `short:"P" long:"password" default:"" description:"Password" env:"MYSQL_PASSWORD"`
 }
 
+type mysqlVersion struct {
+	major     int
+	minor     int
+	patch     int
+	stability string
```
astj, line 29: Is that really means stability?

For example in MySQL 5.5 in Debian, I got following result:
```
mysql> show variables like 'version';
+---------------+---------------------+
| Variable_name | Value               |
+---------------+---------------------+
| version       | 5.5.44-0+deb8u1-log |
+---------------+---------------------+
1 row in set (0.00 sec)
```

Apparently `deb8u1` does not indicate stability.
I guess this is just set by `-DMYSQL_SERVER_SUFFIX` option on build : [ref](https://salsa.debian.org/mariadb-team/mysql/-/blob/mysql-8.0/debian/master/debian/rules#L94). Is it just be treated as "suffix", I think.

check-mysql/lib/check-mysql.go
```diff
@@ -52,12 +62,60 @@ SubCommands:`)
...
+		Passwd:               m.Pass,
+		Net:                  proto,
+		Addr:                 target,
+		AllowNativePasswords: true,
+	}
+
+	return sql.Open("mysql", cfg.FormatDSN())
+}
+
+func getMySQLVersion(db *sql.DB) (*mysqlVersion, error) {
```
astj, line 83: [offtopic]
Maybe the best way to retrieve MySQL version is [`mysql_get_server_version()` C API](https://dev.mysql.com/doc/c-api/8.0/en/mysql-get-server-version.html), but maybe it's not available in go-sql-driver/mysql, maybe…

check-mysql/lib/replication.go
```diff
@@ -22,38 +66,65 @@ func checkReplication(args []string) *checkers.Checker {
...
+	mySQLVersion, err := getMySQLVersion(db)
 	if err != nil {
-		return checkers.Unknown("couldn't execute query")
+		return checkers.Unknown(fmt.Sprintf("Couldn't get MySQL Version: %s", err))
 	}
 
-	if len(rows) == 0 {
-		return checkers.Ok("MySQL is not slave")
+	// MySQL > 8.0.22 supports `SHOW REPLICA STATUS`
+	replicaSupport := !(mySQLVersion.major < 8 || (mySQLVersion.major == 8 && mySQLVersion.minor == 0 && mySQLVersion.patch < 22))
```
astj, line 81: Does it work properly on MariaDB instances?

check-mysql/lib/check-mysql.go
```diff
@@ -52,12 +62,60 @@ SubCommands:`)
...
 		proto = "unix"
 		target = m.Socket
 	}
-	return mysql.New(proto, "", target, m.User, m.Pass, "")
+	cfg := &mysql.Config{
+		User:                 m.User,
+		Passwd:               m.Pass,
+		Net:                  proto,
+		Addr:                 target,
+		AllowNativePasswords: true,
```
astj, line 77: Hmm, do we need to support `AllowOldPasswords` or such? 🤔 (ref: https://pkg.go.dev/github.com/go-sql-driver/mysql#Config )

check-mysql/lib/check-mysql.go
```diff
@@ -1,14 +1,17 @@
...
+	"strconv"
 	"strings"
 
+	"github.com/go-sql-driver/mysql"
 	"github.com/mackerelio/checkers"
-	"github.com/ziutek/mymysql/mysql"
+
 	// MySQL Driver
-	_ "github.com/ziutek/mymysql/native"
+	_ "github.com/go-sql-driver/mysql"
```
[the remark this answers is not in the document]
nonylene: 2009475f3ee649c817c55b4b759a94f72a270d69

check-mysql/lib/check-mysql.go
nonylene: Yes, go-sql-driver/mysql does not have such API.

check-mysql/lib/check-mysql.go
nonylene: I will delete the field since this is not used from check-mysql.

Reason for the name `stability` is below, as an aside.

---

MySQL doc says that release versions is composed of three numbers and an optional suffix, and the suffix is described as `stability level of the release`.

> Release names can also include a suffix to indicate the stability level of the release. Releases within a series progress through a set of suffixes to indicate how the stability level improves. The possible suffixes are: ...
> 
> https://dev.mysql.com/doc/refman/8.0/en/which-version.html

check-mysql/lib/check-mysql.go
nonylene: 68baa67

astj (approved):

check-mysql/lib/replication.go
nonylene: MariaDB is not considered on this PR.

check-mysql/lib/check-mysql.go
nonylene: `old_password` is not supported because the protocol is too legacy (~2004)  (https://github.com/go-sql-driver/mysql/wiki/old_passwords).

nonylene:
Thanks!

PR #492, opened 2021-06-04
[check-disk] upgrade gopsutil to v3 and fix treatment for fstype=none

susisu:
Changes in gopsutil@v3 are listed here: https://github.com/shirou/gopsutil/blob/master/_tools/v3migration/v3Changes.md

- Fixes build failure on macOS https://github.com/shirou/gopsutil/issues/958
- Also fixes treatment for fstype=none

### DIFF
--- a/check-disk/lib/check-disk.go
+++ b/check-disk/lib/check-disk.go
@@ -273,7 +273,7 @@ func listPartitions() ([]gpud.PartitionStat, error) {
 			"ignore":
 			continue
 		case "none":
-			if !strings.Contains(p.Opts, "bind") {
+			if !isBindMount(p.Opts) {
 				partitions = append(partitions, p)
 			}
 		default:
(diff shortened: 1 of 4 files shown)


check-disk/lib/check-disk.go
```diff
@@ -273,9 +273,10 @@ func listPartitions() ([]gpud.PartitionStat, error) {
 			"ignore":
 			continue
 		case "none":
-			if !strings.Contains(p.Opts, "bind") {
-				partitions = append(partitions, p)
+			if !isBindMount(p.Opts) {
+				continue
 			}
+			partitions = append(partitions, p)
```
susisu, line 279: A mountpoint of fstype=none should be included if it is bind-mounted; otherwise ignored.

Ref: https://github.com/coreutils/gnulib/blob/c45faf7f48bcbe57df3bbf8df81a016c5f15df55/lib/mountlist.c#L182-L189
(This is the referenced source that the original code was introduced https://github.com/mackerelio/go-check-plugins/commit/4dce775e0c0d83583d25cc0cb699d2809324d81e)

susisu: Maybe we don't need this at all?
https://github.com/coreutils/gnulib/blob/a742bdb3fd6f5ebaa730824eda7fcfe7faedb451/lib/mountlist.c#L541-L545

lufia: In my opinion it might not be needed in case of reading files provided by procfs. But I think it is good to keep this check if you worry.

lufia (approved):

susisu:
Thanks!

PR #495, opened 2021-06-10
[ci] run tests on the workflow

lufia:
I changed workflow to run tests (**&lt;plugin&gt;/test.sh**) for each plugins.

The "make lint cover testconvention test" task ends up with the reports of each tests of the plugin.

```
OK: check-disk/test.sh
OK: check-ldap/test.sh
OK: check-postgresql/test.sh
OK: check-redis/test.sh
```

see also mackerelio/mackerel-agent-plugins#771


yseto (approved):
LGTM!

PR #498, opened 2021-06-18
[ci]rewrite check-memcached tests. used docker.

yseto:
- check-memcached couldn't test it without the actual command.
- the test was skipped in GitHub Actions, so I replaced it with a test using Docker.


lufia (approved):

yseto:
Thanks!

PR #516, opened 2021-09-28
[check-redis] migrate redis client library to redigo

pyto86pri:
The redis client library `fzzy/radix` seems to be removed, so I migrated the library to `gomodule/redigo`.


susisu (approved):
LGTM!

PR #552, opened 2021-12-17
check-mysql: adding my.cnf config file option with profile

GermanG:
For the use case of having the credentials and connection definition in a my.cnf style file.

### DIFF
--- a/check-mysql/lib/check-mysql.go
+++ b/check-mysql/lib/check-mysql.go
@@ -8,16 +8,19 @@ import (
 	"os"
 	"strings"
 
+	"github.com/go-ini/ini"
 	"github.com/go-sql-driver/mysql"
 	"github.com/mackerelio/checkers"
 )
 
 type mysqlSetting struct {
-	Host   string `short:"H" long:"host" default:"localhost" description:"Hostname"`
-	Port   string `short:"p" long:"port" default:"3306" description:"Port"`
-	Socket string `short:"S" long:"socket" default:"" description:"Path to unix socket"`
-	User   string `short:"u" long:"user" default:"root" description:"Username"`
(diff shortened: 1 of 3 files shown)


check-mysql/lib/check-mysql.go
```diff
@@ -37,6 +40,44 @@ var commands = map[string](func([]string) *checkers.Checker){
...
+				}
+				if s.Key("socket").Value() != "" {
+					m.Socket = s.Key("socket").Value()
+				}
+				if s.Key("user").Value() != "" {
+					m.User = s.Key("user").Value()
+				}
+				if s.Key("password").Value() != "" {
+					m.Pass = s.Key("password").Value()
+				}
```
yseto, line 72: If the configuration file and the argument options are listed together, I want the argument options to take precedence.

yseto (commented):
Thank you for this pull request.

check-mysql/lib/check-mysql.go
GermanG: Makes a lot of sense!

yseto: We will do our best to accept the changes.
If the work seems to be difficult, it would be helpful if you could write a test of the behavior added this time.

yseto: @GermanG 
Happy New Year!
What happened to this pull request?

GermanG: Thanks, same to you!
No news so far. I didn't like my first approach to this, I'm reworking it.

yseto: > I'm reworking it.

sure.
I think it would be better if the values are applied in the order of default, my.cnf reference, and arguments.

yseto:
This issue has been open for some time, but due to lack of activity, I am closing it.
Please reopen if necessary.

PR #583, opened 2022-03-25
[check-aws-cloudwatch-logs] use FilterLogEventsPages API

pyto86pri:
Now check-aws-cloudwatch-logs plugin uses `FilterLogEvents` API and implements pagination by itself. But I think it's better to use `FilterLogEventsPages` API.

### DIFF
--- a/check-aws-cloudwatch-logs/lib/check-aws-cloudwatch-logs.go
+++ b/check-aws-cloudwatch-logs/lib/check-aws-cloudwatch-logs.go
@@ -114,6 +114,7 @@ func (p *awsCloudwatchLogsPlugin) collect(now time.Time) ([]string, error) {
 	}
 	if s.StartTime == nil || *s.StartTime <= now.Add(-time.Hour).Unix()*1000 {
 		s.StartTime = aws.Int64(now.Add(-1*time.Minute).Unix() * 1000)
+		s.NextToken = nil
 	}
 	var messages []string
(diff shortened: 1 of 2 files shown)


check-aws-cloudwatch-logs/lib/check-aws-cloudwatch-logs.go
```diff
@@ -107,66 +107,58 @@ type logState struct {
...
-		}
+func (p *awsCloudwatchLogsPlugin) collect(now time.Time) ([]string, error) {
+	s, err := p.loadState()
+	if err != nil {
+		return nil, err
 	}
-	if startTime == nil {
-		startTime = aws.Int64(time.Now().Add(-1*time.Minute).Unix() * 1000)
+	if s.StartTime == nil || *s.StartTime <= now.Add(-time.Hour).Unix()*1000 {
+		s.StartTime = aws.Int64(now.Add(-1*time.Minute).Unix() * 1000)
```
susisu, line 116: Maybe we should set `s.NextToken = nil` explicitly for this case?

susisu (approved):
LGTM

check-aws-cloudwatch-logs/lib/check-aws-cloudwatch-logs.go
pyto86pri: You're right. I'll fix it!

PR #622, opened 2022-07-15
check-http: add no-restime-success-msg option

sugy:
- Do not output response time on success.
- Omissioning success report in mackerel-agent.
  ref. https://github.com/mackerelio/mackerel-agent/blob/1e7a38a/command/command.go#L465-L468


sugy:
I solved this implementation by wrapping the check command in a shell script. :bow:

PR #629, opened 2022-07-28
Improve tests for check-mysql

susisu:
- Specify docker platform explicitly so that they can be run on Docker Desktop for Mac on Apple silicon
- Run all tests on CI
  - Because of this, it takes about 1 minute longer than before (but I think it's acceptable)
- Fix some unstable tests
- Refactor test scripts

### DIFF
--- a/check-mysql/test.sh
+++ b/check-mysql/test.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+set -ex
+
+./test_57/test.sh
+./test_8/test.sh
--- a/check-mysql/test.sh
+++ b/check-mysql/test.sh
@@ -8,3 +8,4 @@ export PATH=$(pwd):$PATH
 
 ./test_57/test.sh
 ./test_8/test.sh
+./test-tls.sh
--- a/check-mysql/test.sh
+++ b/check-mysql/test.sh
@@ -2,5 +2,9 @@
 
 set -ex
 
+cd "$(dirname "$0")"
+
+export PATH=$(pwd):$PATH
+
 ./test_57/test.sh
 ./test_8/test.sh
--- a/check-mysql/test_57/test.sh
+++ b/check-mysql/test_57/test.sh
@@ -40,7 +40,7 @@ fi
 
 sleep 2
 
-if ! $plugin uptime --host=127.0.0.1 --port=$primary_port --user=$user --password=$password --critical=2 --warning=1; then
+if ! $plugin uptime --host=127.0.0.1 --port=$primary_port --user=$user --password=$password --critical=1 --warning=2; then
 	echo 'FAIL: uptime should be OK'
 	exit 1
 fi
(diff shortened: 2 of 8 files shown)


test.bash
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # prepare tests
-for f in check-*/test.sh check-*/test-*.sh
```
susisu, line 4: There's no `test-*.sh` to be executed directly anymore

check-mysql/test_57/test.sh
```diff
@@ -30,15 +29,17 @@ if $plugin connection --host=127.0.0.1 --port=$primary_port --user=$user --passw
...
-sleep $sleep_time
+
+USER=$user PASSWORD=$password PORT=$primary_port ../wait.sh
 
 if ! $plugin connection --host=127.0.0.1 --port=$primary_port --user=$user --password=$password; then
 	echo 'FAIL: connection should be OK'
 	exit 1
 fi
 
+sleep 2
```
susisu, line 41: Since the following test requires uptime to be > 2, it might be unstable.

check-mysql/test_57/test.sh
```diff
@@ -12,13 +12,12 @@ fi
 
 cd $(dirname $0)
 plugin=$(basename $(realpath ../))
-if ! which -s $plugin
```
susisu, line 15: `which -s` do not work on CI

check-mysql/test_57/test.sh
```diff
@@ -30,15 +29,17 @@ if $plugin connection --host=127.0.0.1 --port=$primary_port --user=$user --passw
 fi
 
 docker-compose up -d
-
 trap 'docker-compose down --rmi local -v' EXIT
-sleep $sleep_time
+
+USER=$user PASSWORD=$password PORT=$primary_port ../wait.sh
```
susisu, line 34: I think it's better than `sleep`.

check-mysql/test_57/test.sh
```diff
@@ -79,6 +80,9 @@ mysql -u$user -p$password --host 127.0.0.1 --port=$replica_port -e """
 START SLAVE USER='repl' PASSWORD='repl';
 """
 
+# starting replication may take a while
+sleep 1
```
susisu, line 84: Because starting replication takes some time, the following test was unstable.

check-mysql/test.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -ex
+
+cd "$(dirname "$0")"
+
+export PATH=$(pwd):$PATH
+
+./test_57/test.sh
+./test_8/test.sh
+./test-tls.sh
```
susisu, line 11: Run sequentially because it's the easiest way to avoid port conflicts.

test.bash
lufia: [may][nit] or else: `check-*/test*.sh`

lufia (approved):
LGTM~

test.bash
susisu: I also wanted to allow each plugin to control how to run its tests, like avoiding port conflicts https://github.com/mackerelio/go-check-plugins/pull/629#discussion_r932153926.
It's still possible if `test.bash` executes all of `test*.sh`, but I think only executing `test.sh` is easier to control.

PR #676, opened 2022-12-22
added compile option, fix packaging format

yseto:
Changed to use xz format for compressing debian packages. refer https://github.com/mackerelio/mackerel-agent-plugins/pull/981

The environment ubuntu-latest for GitHub Actions is ubuntu 22.04.
It turns out that glibc is new and the built binary compatibility is broken in some environments.
So I made a change to build with CGO_ENABLED=0


Arthur1 (approved):
lgtm

yseto:
Thanks for your review.

PR #692, opened 2023-01-20
[check-http] add test.sh

lufia:
I added **check-http/test.sh**. It tests:

- access local HTTP service directly
- access local HTTP service via local squid proxy


yseto (approved):
seems good.

PR #720, opened 2023-02-22
fix gosimple, ineffassign

wafuwafu13:


check-masterha/lib/check_masterha_status.go
```diff
@@ -29,7 +29,7 @@ func (c statusChecker) MakeCommandArgs() []string {
 
 func (c statusChecker) Parse(out string) (checkers.Status, string) {
 	lines := strings.Split(out, "\n")
-	errors := make([]string, 0, 0)
+	errors := make([]string, 0)
```
lufia, line 32: more simply?

```
var errors []string
```

lufia (approved):
thanks!

PR #764, opened 2023-08-07
Remove old rpm packaging

yseto:


lufia (approved):
Announcement
https://mackerel.io/blog/entry/weekly/20230809#We-ended-support-for-Amazon-Linux-AMI-also-known-as-Amazon-Linux-1

PR #766, opened 2023-09-13
support overwrite status

yseto:


lufia (commented):
It is better to update README.md

lufia (approved):
status is one of ok, warning, critical or unknown.

yseto:
Thanks for your review.

PR #767, opened 2023-09-20
[check-tcp] Supports option to monitor that ports are closed.

tukaelu:
TCP connection monitoring is provided by a mechanism that assumes the port is open.

In some cases, it is possible that the state in which the port is closed is correct, and in such cases, the following shell workaround can be used.

```toml
[plugin.checks.dummy]
command = "check-tcp -H <host> -p <port> && exit 2 || exit 0"
```

However, since it is more convenient to be provided as an option, we have added a --expect-closed flag to determine that the state in which the port is closed is normal.

When this option is enabled, it will only work for processes that make connections.


ne-sachirou (approved):
🙆‍♀️

PR #782, opened 2023-09-25
check-ntservice: exclude should be -x, not -E

kmuto:
I found a mistype option in README.md of check-ntservice.
`-E` should be `-x`. (I checked the code and its behavior)

### DIFF
--- a/check-ntservice/README.md
+++ b/check-ntservice/README.md
@@ -43,7 +43,7 @@ command = ["check-ntservice", "--service-name", "SERVICE_NAME"]
 
 ```
   -s, --service-name=    matches if contained in service name
-  -E, --exclude-service= exclude if contained in service name. This option takes precedence over --service-name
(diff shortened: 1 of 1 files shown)


check-ntservice/README.md
```diff
@@ -43,7 +43,7 @@ command = ["check-ntservice", "--service-name", "SERVICE_NAME"]
 
 ```
   -s, --service-name=    matches if contained in service name
-  -E, --exclude-service= exclude if contained in service name. This option takes precedence over --service-name
+  -x, --exclude-service= exclude if contained in service name. This option takes precedence over --service-name
```
Arthur1, line 46: implementation code:

https://github.com/mackerelio/go-check-plugins/blob/f98a267cbb6e5a48c535bd6970650dbcbb791681/check-ntservice/lib/check_ntservice.go#L14

Arthur1 (approved):
👍

PR #788, opened 2023-10-16
use mackerelio/workflows and upgrade maximum Go version to 1.21

lufia:
* use mackerelio/workflows
* upgrade Go


yseto (approved):
seems good.

PR #800, opened 2023-11-14
CGO_ENABLED=0 when build for packaging

Arthur1:
We build the binary for rpm/deb package on `ubuntu-latest` GitHub Actions runner image.

The generated binary may depend on a libc that is too new for the environment we support.

So, build with `CGO_ENABLED=0` flag.


yseto (approved):
seems good.

PR #823, opened 2024-02-22
Reduce check-log errors when a file in the log directory has been removed at the moment of running check-log

ne-sachirou:
2 similer problems.

1. `os.ReadDir` function may return an error from [lstat](https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/os/dir_unix.go;l=132) when the file is removed within the period of running `os.ReadDir` function. If the error occurs, `os.ReadDir` function returns entries it was able to read before the error https://pkg.go.dev/os@go1.22.0#ReadDir .
2. `DirEntry.Info` method may return an error when the file is removed between the process runs `os.ReadDir` function and runs `DirEntry.Info` method. The error satisfies `errors.Is(err, ErrNotExist)` accroding to https://pkg.go.dev/io/fs@go1.22.0#DirEntry .

### DIFF
--- a/check-log/lib/check-log.go
+++ b/check-log/lib/check-log.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"crypto/md5"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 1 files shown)


yseto (approved):
seems good.

PR #857, opened 2024-05-30
[check-mailq] fix pattern

lufia:
This patch is ported from https://github.com/mackerelio/mackerel-agent-plugins/pull/1186


yseto (approved):
seems good.

PR #865, opened 2024-06-10
return CRITICAL instead of UNKNOWN when check-redis reachable is failed

kmuto:
I believe `check-redis reachable` is intented to check a rechability.
Unlike the expectation of returning OK / CRITICAL, this command returns OK / UNKNOWN.

Here is a proposal patch to change the behavior. This may be breaking change, but it should meet user's expectation.


yseto (approved):
seems good.

